PR #50, opened 2015-01-08
Add OpenStack host model doc

Ladas:
Adding openstack host model doc, describing what collumns
will be filled by attributes available from OpenStack Ironic
service.


blomquisg:
This is great!  I'm sure others can fill in more details re: VmWare and other providers.  But, this is fantastic!

PR #55, opened 2015-02-11
Add Rails cops to rubocop

Fryguy:


Fryguy:
Related to ManageIQ/manageiq#1633

PR #58, opened 2015-02-11
Add in Mac Setup

brandondunne:


brandondunne:
cc @jrafanie @Fryguy

developer_setup.md
```diff
@@ -34,6 +34,37 @@
       ```
 
 ### Mac
+  1. Install PostgreSQL
+    - Download and install the .dmg for 9.2.4 for MacOSX:  http://www.enterprisedb.com/products-services-training/pgdownload#osx
+    - Open the dmg and run the installer (It will update your shared memory settings and prompt for a reboot, reboot it)
+    - Open the dmg and run the installer again after the reboot.
```
jrafanie: Does anyone use the dmg way anymore?  @kbrock @Fryguy @matthewd @tenderlove 

I'm using postgresql from homebrew I think.

brandondunne: @jrafanie I didn't write any of this, not sure what you're supposed to do for Mac.  I'm just consolidating unmaintained doc from other sources. :dragon:

Fryguy: Yeah the homebrew way is easier to describe.  I can push a change here with updated Mac doco if necessary

tenderlove: I'm using PG from homebrew as well.

brandondunne: @Fryguy It would be great if you could add the Mac instructions here or in a follow-up PR, whichever one is easier.

kbrock: Think we're down to ~3 devs with Mac postgres (not homebrew).

I encourage everyone to go over to homebrew.
All new installs are using homebrew.

Fryguy:
I updated @brandondunne's fork with a PR to update the Mac instructions.  I changed formatting a bit there, so there is an inconsistency.  Once we merge this one, I'll make a separate PR to fix the formatting and other cleanup in the rest of the document.

PR #59, opened 2015-02-16
Membership can only be checked if you're in the organization.

jrafanie:
Followup to #56 

cc @brandondunne @Fryguy


brandondunne:
:+1:

PR #61, opened 2015-03-03
Add Methods Available for Automation Guide

kallimachos:
Removed downstream content and converted to AsciiDoc for upstream.

### DIFF
--- a/Methods_Available_for_Automation/Methods_Available_for_Automation.adoc
+++ b/Methods_Available_for_Automation/Methods_Available_for_Automation.adoc
@@ -8,11 +8,10 @@ Methods Available for Automation
 // asciidoc -a toc -n -a toclevels=3 Methods_Available_for_Automation.ad
 
 .Abstract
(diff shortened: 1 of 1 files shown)


mfeifer:
Great to see this.  Need to replace all CloudForms Management Engine/CFME instances (as shown in abstract) with ManageIQ.
Also, I thought that this was going in manageiq/manageiq_docs repo.  Guides repo is for developer doc, but then again, this might be better here.  Willing to discuss @Fryguy @xav-rh John Hardy?

kallimachos:
@mfeifer, I think you are right about the repos.  ManageIQ/guides looks like it is aimed at people who want to develop ManageIQ rather than use it, so the Methods guide should probably be in ManageIQ/manageiq_docs.
At any rate, I've replaced all CloudForms Management Engine/CFME instances with ManageIQ.  I'll submit a new pull request with the changes when we confirm which repo the guide should live in.

Fryguy:
This belongs in https://github.com/ManageIQ/manageiq_docs

PR #66, opened 2015-03-12
Use rake evm:compile_assets.

roliveri:
Updated scripts to use evm:compile_assets instead of rake assets:precompile.


roliveri:
@Fryguy please review.

PR #68, opened 2015-04-29
Update to the external authentication guide to mention the new --ipadomain option.

abellotti:
Update to the external authentication guide to mention the new --ipadomain when installing IPA from the CLI.


abellotti:
Thanks @mfeifer, updated and pushed.

PR #70, opened 2015-05-07
Store generated data locally on the appliance.

roliveri:
Modified scripts to prevent generated data from being stored on
the MAC via shared folders.


roliveri:
@jrafanie please review.

PR #71, opened 2015-06-19
Add Fedora 22 setup.

imtayadeway:
* fedora's package manager has changed from yum -> dnf
* nokogiri requires that the patch package be installed separately

### DIFF
--- a/developer_setup.md
+++ b/developer_setup.md
@@ -31,6 +31,35 @@
   su postgres -c "psql -c \"CREATE ROLE root SUPERUSER LOGIN PASSWORD 'smartvm'\""
   ```
 
+#### Fedora 22
+
+* Install packages
+
+  ```bash
+  sudo dnf -y install git-all                            # Git and components
+  sudo dnf -y install memcached                          # Memcached for the session store
+  sudo dnf -y install postgresql-devel postgresql-server # PostgreSQL Database server and to build 'pg' Gem
+  sudo dnf -y install libxml2-devel libxslt-devel patch  # For Nokogiri Gem
+  sudo dnf -y install gcc-c++                            # For event-machine Gem
+  ```
+
+* Enable Memcached
+
+  ```bash
+  sudo systemctl enable memcached
+  sudo systemctl start memcached
+  ```
+
+* Configure PostgreSQL
+
+  ```bash
+  sudo passwd postgres <new_password>
+  su postgres -c 'initdb -D /var/lib/pgsql/data'
+  sudo systemctl enable postgresql
+  sudo systemctl start postgresql
+  su postgres -c "psql -c \"CREATE ROLE root SUPERUSER LOGIN PASSWORD 'smartvm'\""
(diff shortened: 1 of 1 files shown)


jrafanie:
:heart:  Thanks!

@movitto @brandondunne @martinpovolny Please review

developer_setup.md
```diff
@@ -31,6 +31,35 @@
...
+  ```
+
+* Configure PostgreSQL
+
+  ```bash
+  sudo passwd postgres <new_password>
+  su postgres -c 'initdb -D /var/lib/pgsql/data'
+  sudo systemctl enable postgresql
+  sudo systemctl start postgresql
+  su postgres -c "psql -c \"CREATE ROLE root SUPERUSER LOGIN PASSWORD 'smartvm'\""
```
Fryguy: This looks identical to Fedora 20 section...can we just update that section title to be Fedora 20+.  Unless I'm missing something.

imtayadeway: yep it is identical - i'll adjust as above

imtayadeway:
@Fryguy i just removed the redunancy and added an extra line for clarity. i'll squash if you think this is OK

movitto:
@jrafanie I run miq off the appliance and haven't set it up on fedora in a while so I can't comment on the specific steps. Generally speaking it looks fine though.

I wonder if the app & underlying stack could simply be yum installed at some point. That + an installer could simplify the setup experience.

brandondunne:
:+1: LGTM

jrafanie:
@movitto rpms aren't cross platform so that makes it harder.  If we used a platform independent packaging format, maybe but it feels like a lot of overhead for what bundler does already for at least 90% of our stack.  

More appliance formats + container support could simplify the process provided each platform makes it easy to setup the hypervisor/docker environment.  If you want to help out in those areas, we can certainly use it!  :heart:

movitto:
@jrafanie appliances / containers would still be limited to the environments we support & would still entail building the multiple environments as you mentioned.

Packages are fairly platform specific but still are needed & provide tradeoffs w/ a bundler centric solution. An alternate solution would be to build packages for multiple platforms as needed.

jrafanie:
Yes, dependencies are hard :cry: especially when we have platform specific solutions to work with.

Fryguy:
To @movitto's point, would it make sense to make some kind of meta-rpm that has all the dependencies for Fedora?  Like a `manageiq-devel` rpm that just has dependencies on those other devel packages?

Fryguy:
Also, I just noticed that the package manager is `dnf` and not `yum`. **That** was the difference.  When did they change that?  Is `yum` even available anymore?

imtayadeway:
@Fryguy you can still use `yum` commands in 22 but it will complain and issue the `dnf` command for you instead.

jrafanie:
:+1: Let's make any further changes as we find issues.  Thanks @imtayadeway

PR #72, opened 2015-06-23
Added --without qpid:metric_fu to bundle install.

roliveri:


roliveri:
@jrafanie please review.

miq-dev-appliance-setup/MAC/on_appliance/miq-setup.sh
```diff
@@ -6,7 +6,7 @@ cd $MIQ_DIR/vmdb || exit 1
 
 # Update gems.
 echo "**** Updating gems..."
-bundle update || exit 1
+bundle install --without qpid:metric_fu || exit 1
```
jrafanie: So, this --without option is a "remembered" option, so when you specify it the first time, it gets saved to the application/.bundle/config file so you shouldn't need it a second time.

miq-dev-appliance-setup/MAC/on_appliance/source-updated.sh
```diff
@@ -38,7 +38,7 @@ cd $MIQ_DIR/vmdb || exit 1
 
 # Update gems.
 echo "**** Updating gems..."
-bundle update || exit 1
+bundle install --without qpid:metric_fu || exit 1
```
jrafanie: I'm not sure if you need this in both places, see above.

I'm fine with it but it's due to change as we upgrade/change things.

jrafanie:
Chatted with @roliveri, decided that it's ok to duplicate this for now even though these values could change as we re-organize things... it's a problem, we can remove the duplication then.

PR #76, opened 2015-07-29
Don't allow GlobalVars by default.

brandondunne:
Gems and other libraries shouldn't use them


brandondunne:
@Fryguy Please review

cc @jrafanie @matthewd

jrafanie:
:+1:

PR #90, opened 2015-12-02
Add some additional hints for OS/X

pilhuhn:
Homebrew installs the database as `postgres`. So creating the `root` user needs to run as the postgres database. The Homebrew user is has enough rights to create the role, but psql will not find a `whoami` database to connect to, while the `postgres`one is always there.

### DIFF
--- a/developer_setup.md
+++ b/developer_setup.md
@@ -83,18 +83,10 @@ As per 20+, with the following changes:
   brew install git
   brew install memcached
   brew install postgresql
-  brew install graphviz
-  ```
-
-If `bundle install` (when running `bin/setup` below) fails in installing the `eventmachine` gem,
-you may want to run the following and then retry `bundle install`:
-
-  ```bash
-  brew link openssl --force # If installation of eventmachine gem fails
   ```
 
 
-* Configure PostgreSQL
+* Configure and start PostgreSQL
 
   ```bash
   # Enable PostgreSQL on boot
@@ -178,3 +170,22 @@ bin/setup
 * Now you can start the full application with `bundle exec rake evm:start`.
   You can access it at <IP_ADDRESS>:3000. Default username is `admin` and password `smartvm`
 * [Running in minimal mode](developer_setup/minimal_mode.md)
+
+
+#### Some troubleshooting notes
+
+* First login fails
+
+Make sure you have memcached running. If not stop the server with `bundle exec rake evm:stop`,
+start memcached and retry.
+
+* OS/X install is failing on  the `eventmachine` gem
+
+If `bundle install` (when running `bin/setup` below) fails in installing the `eventmachine` gem,
+you may want to run the following and then retry `bundle install`:
+
+  ```bash
+  brew link openssl --force # If installation of eventmachine gem fails
+  ```
+
+
--- a/developer_setup.md
+++ b/developer_setup.md
@@ -83,7 +83,16 @@ As per 20+, with the following changes:
   brew install git
   brew install memcached
   brew install postgresql
+  brew install graphviz
   ```
+
+If `bundle install` (when running `bin/setup` below) fails in installing the `eventmachine` gem,
+you may want to run the following and then retry `bundle install`:
+
+  ```bash
+  brew link openssl --force # If installation of eventmachine gem fails
+  ```
+
 
 * Configure PostgreSQL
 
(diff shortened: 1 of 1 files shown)


kbrock:
added a comment

kbrock:
I'm mixed on force linking openssl.
Maybe move the advice to a bullet note down in the "Get the Rails environment up and running" section?

```
otool -L /Users/kbrock/.gem/ruby/2.2.3/gems/eventmachine-1.0.8/ext/rubyeventmachine.bundle
	/usr/lib/libssl.0.9.8.dylib (compatibility version 0.9.8, current version 0.9.8)
	/usr/lib/libcrypto.0.9.8.dylib (compatibility version 0.9.8, current version 0.9.8)
	/usr/lib/libz.1.dylib (compatibility version 1.0.0, current version 1.2.5)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1213.0.0)
	/usr/local/lib/libgmp.10.dylib (compatibility version 13.0.0, current version 13.0.0)
	/usr/lib/libobjc.A.dylib (compatibility version 1.0.0, current version 228.0.0)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 120.0.0)
```

pilhuhn:
Without the forcelink eventmachine failed to install and it looks like virtually everyone who had issues with it used the forcelink. But there may be better options that I am not aware of.

kbrock:
no, forcelink is good.

I've had to do that for openssl in the past.
But the `--force` is there because homebrew does advise against it.
And that is potentially invasive for other stuff.

So I like having it in there. Just would prefer it as more of a 'if you have a problem' rather than 'you should do this preemptively'

Fryguy:
Why is graphviz needed?

pilhuhn:
The ruby graphviz gem, that is installed by bundle install prints that it 
needs graphviz and that I should install it (manually).


Am 2. Dezember 2015 10:36:36 nachm. schrieb Jason Frey 
<notifications@github.com>:

pilhuhn:
Btw: should copying of the developer cert explicitly be named here?

kbrock:
@pilhuhn sure, sounds good.

pilhuhn:
@kbrock Ok. I've also added option `-w` to `launchctl load`, so that the services start on reboot (otherwise os/x would register them with launchd, but not start them).

kbrock:
also keep in mind that `bin/setup` is your friend and has many of these steps ==> #92

kbrock:
sorry, the key thing messed this up - could you rebase?

pilhuhn:
@kbrock done, squashed and reworded the note about linking the ssl to follow the bin/setup change.

developer_setup.md
```diff
@@ -83,17 +83,34 @@ As per 20+, with the following changes:
...
+
+
 * Configure PostgreSQL
 
   ```bash
   # Enable PostgreSQL on boot
   mkdir -p ~/Library/LaunchAgents
   ln -sfv /usr/local/opt/postgresql/*.plist ~/Library/LaunchAgents
-  launchctl load ~/Library/LaunchAgents/homebrew.mxcl.postgresql.plist
+  launchctl load -w ~/Library/LaunchAgents/homebrew.mxcl.postgresql.plist
```
kbrock: +1 nice

developer_setup.md
```diff
@@ -83,17 +83,34 @@ As per 20+, with the following changes:
...
 
-  psql -c "CREATE ROLE root SUPERUSER LOGIN PASSWORD 'smartvm'"
+  psql -d postgres -c "CREATE ROLE root SUPERUSER LOGIN PASSWORD 'smartvm'"
```
kbrock: +1 nice

developer_setup.md
```diff
@@ -83,17 +83,34 @@ As per 20+, with the following changes:
...
+  ```
+
+* Start memcached
+
+  ```bash
+  # Enable Memcached on boot
+  ln -sfv /usr/local/opt/memcached/homebrew.mxcl.memcached.plist ~/Library/LaunchAgents
+  launchctl load -w ~/Library/LaunchAgents/homebrew.mxcl.memcached.plist
```
kbrock: +1

developer_setup.md
```diff
@@ -83,17 +83,34 @@ As per 20+, with the following changes:
   brew install git
   brew install memcached
   brew install postgresql
+  brew install graphviz
   ```
 
+If `bundle install` (when running `bin/setup` below) fails in installing the `eventmachine` gem,
```
kbrock: Would you consider moving this to below the `bin/setup` step? If people have a problem with that, they will look at notes in that section, not up here.

Then it reads something like

"If bin/setup fails on installing the `evenmachine` gem, run the following command and then retry bundle install:

developer_setup.md
```diff
@@ -83,17 +83,34 @@ As per 20+, with the following changes:
   brew install git
   brew install memcached
   brew install postgresql
+  brew install graphviz
```
kbrock: I know this helps bundler in printing their pretty graphs, but it is optional and not really necessary for our product

developer_setup.md
pilhuhn: While I agree with you, I left it in the OS/X section as it is OS/X specific. The `bin/setup` step is generic.

But then it could be "If bin/setup fails on OS/X on installing the evenmachine gem, run the following command and then retry bundle install:"

pilhuhn:
@kbrock removed graphviz and moved the os/x hint down to a general troubleshooting section. This also mentions not having memcached, which seems to be a common issue

PR #92, opened 2015-12-03
Recommend the new bin/setup script

### LINKED ISSUE #91: guides/developer_setup doesn't work as-in (Fedora 23)
I couldn't start the installation with: 
gem install bundler -v "~>1.7.4"

but it worked with:
gem install bundler -v "~>1.8.4"
(not sure what exact minimal version is necessary)

Then the "bundle install"  command would fail on nokogiri-1.6.7 build, to pass that I had to do:
sudo dnf install rpm-build

matthewd:
I noticed in https://github.com/ManageIQ/guides/pull/90 that we're currently listing manual steps for which we have helpful automation.

Helpful automation is helpful.

Fixes #91.

### DIFF
--- a/developer_setup.md
+++ b/developer_setup.md
@@ -155,11 +155,7 @@ git fetch other_user
 ### Get the Rails environment up and running
 
 ```bash
-gem install bundler -v "~>1.7.4"
-bundle install
-cp config/database.pg.yml config/database.yml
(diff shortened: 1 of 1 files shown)


kbrock:
nice. I was just about to comment on that pr

PR #93, opened 2015-12-04
OSX/Hombrew - Use default db for role creation

chrisarcand:
I don't recall if this is true for most package managers, but Homebrew
does not create a database for the user's name; Thus, when you run
`psql` or the command here to add the smartvm role, you'll get an error
saying that the database <username> doesn't exist.

Changes the command here to just connect to the internal postgres db to
run this command so it always works exactly as intended.


chrisarcand:
Whoops. Looks like this is already covered in https://github.com/ManageIQ/guides/pull/90 :+1:

PR #94, opened 2015-12-08
Test suite doc; tweaks

chrisarcand:
I've tweaked the server startup doc to be a bit more clear and added a doc page on running the test suites.

Suggestions welcome, especially for the Self Service setup as I just made that up quickly without looking too closely at the runner.

### DIFF
--- a/developer_setup.md
+++ b/developer_setup.md
@@ -164,13 +164,15 @@ git fetch other_user
 ### Get the Rails environment up and running
 
 ```bash
-bin/setup
+bin/setup                  # Installs dependencies, config, prepares database, etc
+bundle exec rake evm:start # Starts the ManageIQ EVM Application in the background
+bundle exec rails s        # Starts the application server
 ```
 
-* Now you can start the full application with `bundle exec rake evm:start`.
-  You can access it at <IP_ADDRESS>:3000. Default username is `admin` and password `smartvm`
-* [Running in minimal mode](developer_setup/minimal_mode.md)
-
+* You can now access the application at `http://localhost:3000`. The default username is `admin` with password `smartvm`.
+* There is also a minimal mode available to start the application with fewer services and workers for faster startup or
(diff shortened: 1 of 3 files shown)


chrisarcand:
cc/ @Fryguy

developer_setup/running_test_suites.md
```diff
@@ -0,0 +1,57 @@
...
+* **javascript**
+* **self_service**
+* **brakeman**
+
+Each suite has a pair of Rake tasks to setup and run them, in the form of `test:<type>` and
+`test:<type>:setup` (if required):
+
+```bash
+rake test:vmdb                   # Run all specs except migrations, replication, and automation
+rake test:vmdb:setup             # Setup environment for vmdb specs
```
kbrock: would it make more sense to list the setup before the running step?
(although, I guess alphabetical makes sense to me too)

developer_setup.md
```diff
@@ -155,9 +155,12 @@ git fetch other_user
 ### Get the Rails environment up and running
 
 ```bash
-bin/setup
+bin/setup                  # Installs dependencies, config, prepares database, etc
+bundle exec rake evm:start # Starts the ManageIQ EVM Application in the background
```
kbrock: do we need to run `evm:start` and rails s?

May want more of a note around configuring fewer resources / spartian stuff here.

developer_setup/running_test_suites.md
chrisarcand: I just pasted in what Rake fed me, but did go out of my way to put the main test suite up top. Sure, makes sense to me :)

developer_setup.md
chrisarcand: Hmm I'm not sure what you mean here; haven't dug in enough to know about those options or what's available. Is this beyond the minimal mode guide referenced below?
I'll add whatever specific things you think would be helpful.

chrisarcand:
Rebased and ready for another look.

PR #98, opened 2016-01-22
Update Fedora Dev setup, deduplicate with CentOS setup

bdunne:
- dnf works in Fedora 20+
- Deduplicate setup for extremely similar platforms

### DIFF
--- a/developer_setup.md
+++ b/developer_setup.md
@@ -2,16 +2,24 @@
 
 ### Install System Packages
 
-#### Fedora 20+
+#### Fedora / CentOS 7
+
+* CentOS only - Enable EPEL & install DNF
+
+  ```bash
+  sudo rpm -ivh https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
+  sudo yum -y install dnf
+  ```
 
 * Install packages
 
   ```bash
-  sudo yum -y install git-all                            # Git and components
-  sudo yum -y install memcached                          # Memcached for the session store
-  sudo yum -y install postgresql-devel postgresql-server # PostgreSQL Database server and to build 'pg' Gem
-  sudo yum -y install libxml2-devel libxslt-devel        # For Nokogiri Gem
-  sudo yum -y install gcc-c++                            # For event-machine Gem
+  sudo dnf -y install git-all                            # Git and components
+  sudo dnf -y install memcached                          # Memcached for the session store
+  sudo dnf -y install postgresql-devel postgresql-server # PostgreSQL Database server and to build 'pg' Gem
+  sudo dnf -y install libxml2-devel libxslt-devel patch  # For Nokogiri Gem
+  sudo dnf -y install gcc-c++                            # For event-machine Gem
+  sudo dnf -y install openssl-devel                      # For rubygems
   ```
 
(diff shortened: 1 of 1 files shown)


bdunne:
@jrafanie @Fryguy Please review. :cookie:

jrafanie:
LGTM :+1:

developer_setup.md
```diff
@@ -43,32 +18,25 @@ As per 20+, with the following changes:
...
+* Configure PostgreSQL
 
   ```bash
+  sudo passwd postgres <new_password>
   sudo postgresql-setup initdb
-  sudo su postgres -c "echo local all all trust > /var/lib/pgsql/data/pg_hba.conf"
-  sudo systemctl start memcached
-  sudo systemctl enable memcached
-  sudo systemctl start postgresql
+  sudo sed -i.bak 's/\(^local\s*\w*\s*\w*\s*\)\(peer$\)/\1trust/' /var/lib/pgsql/data/pg_hba.conf
```
chrisarcand: Are we sure we shouldn't do `md5` here?

developer_setup.md
```diff
@@ -2,39 +2,14 @@
...
-* Configure PostgreSQL
+* Enable EPEL (CentOS only)
 
   ```bash
-  sudo passwd postgres <new_password>
-  su postgres -c 'initdb -D /var/lib/pgsql/data'
-  sudo systemctl enable postgresql
-  sudo systemctl start postgresql
-  su postgres -c "psql -c \"CREATE ROLE root SUPERUSER LOGIN PASSWORD 'smartvm'\""
+  sudo rpm -ivh https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
```
agrare: No dnf in centos as of 7.2.1511 (Core) but it can be installed from EPEL, maybe say `Enable EPEL and install DNF (CentOS only)`?

agrare:
:+1: for fedora, just the dnf/yum issue for me on centos7

developer_setup.md
bdunne: `md5` is probably a good idea.  I just consolidated what was there, I'll update it.

developer_setup.md
bdunne: Ok, I'll check it out.

developer_setup.md
kbrock: For our actual appliances, for the local unix file socket we use trust. If the user has access to the local machine. Then they probably can reconfigure the database how ever they want.
For network transfer (even from 127.0.0.1) we use md5 passwords.

Are we sure that the local line is in the text file?

```bash
grep -q '^local\s' pg_hba.conf || echo "local all all trust" > pg_hba.conf
sudo sed -i.bak 's/\(^local\s*\w*\s*\w*\s*\)\(peer$\)/\1trust/' pg_hba.conf
```

bdunne:
@jrafanie @agrare @kbrock @chrisarcand Updated based on your comments.

agrare:
:+1: LGTM

jrafanie:
LGTM, @kbrock Please merge if you're happy with the pg_hba regex magic.

chrisarcand:
:+1: :+1:

PR #102, opened 2016-02-14
developer_setup.md: fix some steps and deps - tested on clean Centos 7.1

cben:
A couple "Configure PostgreSQL" commands didn't work as-is; clarified where password is asked.

rbenv install 2.2.0 gives an obscure linker warning fixed by having `libffi-devel` [https://github.com/rbenv/ruby-build/issues/690].
Their FAQ actually suggests more [https://github.com/rbenv/ruby-build/wiki#suggested-build-environment]:

    dnf install -y gcc bzip2 openssl-devel libyaml-devel libffi-devel readline-devel zlib-devel gdbm-devel ncurses-devel

but `bzip2 libffi-devel readline-devel` (plus `openssl-devel` we already listed) were enough to build.
(not sure if these are relevant for rvm/ruby-install, or for Ruby 2.2.1+, but as long as guide says "required version is 2.2", it should work with 2.2.0)

nodejs is the obvious option to satisfy ExecJS, otherwise sprockets/es6 gem refuses to load.
(also nodejs + npm needed for self-service tests - but these are separate repo now, right?)

### DIFF
--- a/developer_setup.md
+++ b/developer_setup.md
@@ -17,8 +17,10 @@
   sudo dnf -y install git-all                            # Git and components
   sudo dnf -y install memcached                          # Memcached for the session store
   sudo dnf -y install postgresql-devel postgresql-server # PostgreSQL Database server and to build 'pg' Gem
+  sudo dnf -y install bzip2 libffi-devel readline-devel  # For rbenv install 2.2.0 (might not be needed with other Ruby setups)
   sudo dnf -y install libxml2-devel libxslt-devel patch  # For Nokogiri Gem
   sudo dnf -y install gcc-c++                            # For event-machine Gem
+  sudo dnf -y install nodejs                             # For ExecJS Gem used by sprockets/es6 Gem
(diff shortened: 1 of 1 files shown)


cben:
BTW, http://manageiq.org/documentation/development/developer_setup/ is already out of date wrt to this file.

matthewd:
> not sure if these are relevant for rvm/ruby-install, or for Ruby 2.2.1+, but as long as guide says "required version is 2.2", it should work with 2.2.0

"2.2" != "2.2.0"; you definitely shouldn't be using 2.2.0 -- more recent 2.2.x releases contain security fixes.

cben:
Fair enough, I'm a total ruby newbie.
Do you want the guide to say something like "Required/recommended ruby version is 2.2.4"?  Or "latest release of 2.2.x"? (consider this guide is not only for core devs, it's also "how do I run this from source" for everyone).

Anyway, re-tested with 2.2.4, works without libffi-devel and with the other changes.

PR #106, opened 2016-02-18
update postgres setup in developer setup notes

abonas:


chessbyte:
@Fryguy @bdunne @carbonin Please review

developer_setup.md
```diff
@@ -37,8 +37,17 @@
   sudo passwd postgres
   # ↑ will ask to choose a password
   sudo postgresql-setup initdb
+  # if the above usage of postgresql-setup fails with 'command not found', use the following command with the relevant postgres 9.x version - below an example for postgres 9.1:
```
carbonin: I think the distinction is whether or not your system is managed by init or systemd as described [here](https://docs.fedoraproject.org/en-US/Fedora/16/html/Release_Notes/sect-Release_Notes-Changes_for_Sysadmin.html#id2838704) rather than whether the command is found or RHEL vs Fedora

So this section can be simplified to:
### For systems using systemd units
`sudo postgresql-setup initdb`
### For systems using SysV init scripts
`sudo service <postgres service name> initdb`

abonas: @carbonin I don't think it has to do with rhel vs fedora or related to systemd.
As described here in the postgres setup wiki, postgresql-setup is not used:
https://wiki.postgresql.org/wiki/YUM_Installation
I also asked a postgres DBA who told me postgresql-setup command was deprecated/removed.

Fryguy: If it's completely deprecated, then line 39 should just go away.

developer_setup.md
```diff
@@ -37,8 +37,17 @@
...
+  # for rhel:
+  sudo service postgresql-9.1 initdb
+  # for fedora:
+  sudo systemctl postgresql-9.1 initdb
```
Fryguy: Our minimum version is PG 9.4, so the example should probably use that  (yes, I know the minimum versions are not listed...we should do that)

developer_setup.md
abonas: I tried to find any official notice on deprecation, but didn't find
anything. And since I am not checking instructions on all operating
systems, I preferred the "caution mode notice"

developer_setup.md
carbonin: Either way `systemctl <service> <action>` is not proper syntax for the command and will always fail.

developer_setup.md
carbonin: Also running `service <postgres service name> initdb` on a current CentOS appliance yields the following output:
```
[root@localhost ~]# service rh-postgresql94-postgresql initdb
Hint: the preferred way to do this is now "/opt/rh/rh-postgresql94/root/usr/bin/postgresql-setup --initdb --unit rh-postgresql94-postgresql"
 * Initializing database in '/var/opt/rh/rh-postgresql94/lib/pgsql/data'
```

bdunne:
@abonas What OS are you using?  This section of the setup guide should work for RHEL 7 / CentOS 7 / Fedora 2*.  The only modification that may be necessary is changing the Postgres service name or data directory, but I think there are too many variations to list in the setup guide.  If we want to go that route, the setup guide should be converted to a script for developers to run.

Maybe we need a section in the setup guide for older operating systems like RHEL 6 (I'm guessing that's what you're using), but you may run into other Application - OS integration issues with that.

abonas:
I'm closing the PR as there are too many variations on the setup, and it seems based on the review comments that adding the current fixes won't be efficient.

PR #109, opened 2016-02-24
Grepping /var/lib/pgsql/data/pg_hba.conf w/o sudo fails with Permission denied

Jiri-Kremser:
Purpose of this command is not to rewrite the `pg_hba.conf` if there is already a record for the local. However, it doesn't work as expected because the grep fails all the time when running without the sudo.

All in all, it kind of works (not as supposed), but it produces this err msg:

```bash
[jkremser@dhcp-10-40-1-43 manageiq master]λ grep -q '^local\s' /var/lib/pgsql/data/pg_hba.conf || echo "local all all trust" | sudo tee /var/lib/pgsql/data/pg_hba.conf
grep: /var/lib/pgsql/data/pg_hba.conf: Permission denied
```

### DIFF
--- a/developer_setup.md
+++ b/developer_setup.md
@@ -38,7 +38,7 @@
   sudo passwd postgres
   # ↑ will ask to choose a password
   sudo postgresql-setup initdb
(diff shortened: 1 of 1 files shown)


chessbyte:
@gtanzillo @carbonin @jrafanie @Fryguy please review

gtanzillo:
:+1: LGTM

Issue #111, opened 2016-02-29
Missing example of Ruby installation for Developer Setup


gbraad:
The document [developer_setup.md](https://github.com/ManageIQ/guides/blob/master/developer_setup.md) mentions the need of Ruby 2.2.2+ under the heading _Install Ruby_, but it does not give an example of what to do in case people are unfamiliar to the use of a version manager.

gbraad:
Fiirst get the requirements to perform a build of a newer Ruby version

```bash
sudo yum install -y git-core zlib zlib-devel gcc-c++ patch readline readline-devel libyaml-devel libffi-devel openssl-devel make bzip2 autoconf automake libtool bison curl sqlite-devel
git clone git://github.com/sstephenson/rbenv.git .rbenv
echo 'export PATH="$PATH:$HOME/.rbenv/bin"' >> ~/.bash_profile
echo 'eval "$(rbenv init -)"' >> ~/.bash_profile
git clone git://github.com/sstephenson/ruby-build.git ~/.rbenv/plugins/ruby-build
echo 'export PATH="$PATH:$HOME/.rbenv/plugins/ruby-build/bin"' >> ~/.bash_profile
source ~/.bash_profile
```

and then use the following to perform an install of Ruby 2.2.4
```
rbenv install -v 2.2.4
rbenv global 2.2.4
```

chessbyte:
Addressed in https://github.com/ManageIQ/guides/pull/199

PR #112, opened 2016-02-29
Add example of Ruby version manager for developer setup documentation

gbraad:
Add an example of `rbenv` to install an updated version of Ruby

See #111

### DIFF
--- a/developer_setup.md
+++ b/developer_setup.md
@@ -88,6 +88,26 @@
   * [rbenv](https://github.com/sstephenson/rbenv)
 * Required Ruby version is 2.2.2+
 
+#### Example using rbenv
+
+First get the requirements to perform a build of a newer Ruby version
+
+```bash
+sudo yum install -y git-core zlib zlib-devel gcc-c++ patch readline readline-devel libyaml-devel libffi-devel openssl-devel make bzip2 autoconf automake libtool bison curl sqlite-devel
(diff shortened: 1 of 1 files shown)


Fryguy:
@bdunne Thoughts?

developer_setup.md
```diff
@@ -88,6 +88,26 @@
   * [rbenv](https://github.com/sstephenson/rbenv)
 * Required Ruby version is 2.2.2+
 
+#### Example using rbenv
+
+First get the requirements to perform a build of a newer Ruby version
+
+```bash
+sudo yum install -y git-core zlib zlib-devel gcc-c++ patch readline readline-devel libyaml-devel libffi-devel openssl-devel make bzip2 autoconf automake libtool bison curl sqlite-devel
```
bdunne: I think that packages belong up in the OS specific section above with an appropriate comment

bdunne:
@gbraad Aside from the packages, I don't see anything that isn't already covered in the `rbenv` link a couple lines above.  I'd prefer not to add documentation for other projects here unless we really need to.  Did you run into a problem that isn't covered under the `rbenv` install guide?

kbrock:
I'd like to agree.
Making sure all the packages are up in the OS section sounds good.
Duplicating the `rbenv` docs seems a bit much.

PR #116, opened 2016-03-08
Add cmake to packages to be installed from brew for rugged gem

pilhuhn:
rugged needs cmake to create the native part and cmake is not installed on os/x by default.


jrafanie:
Thanks @pilhuhn :+1:

PR #121, opened 2016-04-12
Install "npm" in CentOS and Fedora

jhernand:
In RPM distributions like CentOS and Fedora the "npm" package is separate, and
not installed by default when Node.js is installed. As a result the execution
of the "bin/setup" will fail if the instructions are followed to setup the
environment from scratch, because it uses the "npm" command to install bower.
This patch changes the instructions so that the "npm" package is also
installed.

Signed-off-by: Juan Hernandez <juan.hernandez@redhat.com>

### DIFF
--- a/developer_setup.md
+++ b/developer_setup.md
@@ -21,7 +21,7 @@
   sudo dnf -y install libxml2-devel libxslt-devel patch  # For Nokogiri Gem
   sudo dnf -y install gcc-c++                            # For event-machine Gem
   sudo dnf -y install sqlite-devel                       # For sqlite3 Gem
-  sudo dnf -y install nodejs                             # For ExecJS Gem used by sprockets/es6 Gem
+  sudo dnf -y install nodejs npm                         # For ExecJS Gem and bower
   sudo dnf -y install openssl-devel                      # For rubygems
   sudo dnf -y install cmake                              # For rugged Gem
   ```


jhernand:
@cben can you review?

developer_setup.md
```diff
@@ -21,7 +21,7 @@
   sudo dnf -y install libxml2-devel libxslt-devel patch  # For Nokogiri Gem
   sudo dnf -y install gcc-c++                            # For event-machine Gem
   sudo dnf -y install sqlite-devel                       # For sqlite3 Gem
-  sudo dnf -y install nodejs                             # For ExecJS Gem used by sprockets/es6 Gem
+  sudo dnf -y install nodejs npm                         # For ExecJS Gem used by sprockets/es6 Gem
```
cben: Can you update the comment to something like "For ExecJS Gem and bower"?

cben:
LGTM :+1: 

For some reason I believed npm comes from nodejs package on fedora, but no it's a separate package on my Fedora 23.
On CentOS 7 I see it's in EPEL: https://dl.fedoraproject.org/pub/epel/7/x86_64/repoview/npm.html so should also work.
[Let me know if you want me to test the whole flow in a VM.]

cben:
(I can only review, don't have merge powers)

jhernand:
Updated the comment, as suggested by @cben

PR #124, opened 2016-04-28
Adds mention about bin/update to dev setup guide

petrblaho:


kbrock:
Hi Petr,

Could you put in some text explaining what updating may mean?

Something to the extend that after fetching the data from `git`, you will want to update the javascript libraries and gem dependencis

Thanks

petrblaho:
@kbrock good idea...

PR #125, opened 2016-05-03
Add rubocop rule for curly braces in hash parameter to depend on context

abonas:
Without this rule, rubocop warns on curly braces in hash parameters, even when all the parameters are hashes, hence removing only from the last one decreases code readability.

Example:
https://github.com/ammendonca/manageiq/blob/c063ed1fea40e8486b90fd26dd879fc9572a3caa/spec/services/middleware_topology_service_spec.rb#L72

More info here:
https://github.com/bbatsov/rubocop/issues/2063

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -44,6 +44,8 @@ SpaceInsideHashLiteralBraces:
   EnforcedStyle: no_space
 TrivialAccessors:
   AllowPredicates: true
+Style/BracesAroundHashParameters:
+  EnforcedStyle: context_dependent
 #
 # Enabled/Disabled
 #


abonas:
cc @ammendonca 
@Fryguy @chessbyte @jrafanie please review. thanks

abonas:
@Fryguy @chessbyte @jrafanie please review. thanks

jrafanie:
I'm good with this change.  For reference, it would be helpful to show `before` and `after rubocop` code example, and WHY it's silly to listen to rubocop ALWAYS.  In other words, why `before` is good and `after rubocop` is either wrong or harder to read.   Many of these things are subjective so a before and after and justification of why the before is better is helpful.  Also, many cop suggestions are better only sometimes... It's helpful to say "this rule's suggestion is worse never, sometimes, most times, or always..." and provide explanation of this.  

These cops are for our own benefit so anything that's always NOT what we want is helpful to fix

abonas:
@jrafanie isn't my comment above providing with the information regarding when does the rule help, as well as a link to a code example where it occurs in miq?  I also put there a link to another explanation (in the other link to rubocop issue)

jrafanie:
It helped a bit but I had to spend a few minutes following links to understand the context.  A summary including before/after and why would have been helpful.  For example, it's not clear what the setting does in terms of various before/after code examples and if/when it would complain.

I still don't know exactly what context_dependent actually does.

```
Style/BracesAroundHashParameters:
   EnforcedStyle: context_dependent
```

This is why I didn't review it immediately.  It wasn't clear to me on the first pass, so I put it in my "I have to go back and review this more in depth" mental backlog.  So, I got distracted and moved onto other mentions with the intent to dig in later.

abonas:
> I still don't know exactly what context_dependent actually does.
> 
> Style/BracesAroundHashParameters:
>    EnforcedStyle: context_dependent

the documentation/code for context_dependent option is here:
https://github.com/bbatsov/rubocop/blob/master/config/default.yml#L268
I hope it can be merged now. thanks

jrafanie:
A more descriptive commit message that explains why, very much like you've said here, would help in case we git blame this commit 6 months from now.

I'm good with merging when this is done.

:+1:

abonas:
@jrafanie please see updated commit message.

jrafanie:
Thanks, that helps explain the "why" much better.

PR #126, opened 2016-05-16
remove confusing instruction to run rails s after evm:start

abonas:
cc @karelhala @mtho11 @ammendonca 
@blomquisg please review

### DIFF
--- a/developer_setup.md
+++ b/developer_setup.md
@@ -156,7 +156,6 @@ git fetch other_user
 ```bash
 bin/setup                  # Installs dependencies, config, prepares database, etc
 bundle exec rake evm:start # Starts the ManageIQ EVM Application in the background
-bundle exec rails s        # Starts the application server
 ```
 
 * You can now access the application at `http://localhost:3000`. The default username is `admin` with password `smartvm`.


blomquisg:
cc @Fryguy, @jrafanie

developer_setup.md
```diff
@@ -156,7 +156,7 @@ git fetch other_user
 ```bash
 bin/setup                  # Installs dependencies, config, prepares database, etc
 bundle exec rake evm:start # Starts the ManageIQ EVM Application in the background
-bundle exec rails s        # Starts the application server
+bundle exec rails s        # Starts the application server - no need to run this if evm:start was executed
```
chrisarcand: While clarification here might be needed, saying this exact thing is bit misleading; The default setup happens to include a UI worker but you could have changed that in your configuration. 

Since this _is_ the intro developer setup, perhaps simply removing `rails s` would be less confusing. Knowing how to turn off your UI worker and instead run it yourself in the console is beyond the scope of this very particular set of instructions.

Fryguy: :+1: Remove the line altogether.

kbrock: But don't most developers use `bundle exec rails s`?

chrisarcand: I don't think that's necessarily true. @blomquisg for example.

abonas: @chrisarcand @Fryguy I will remove this line. the PR is initiated due to a fact that developers who are new to miq and new to rails run both evm:start and rails s, because this is how it sounds from the guide.
@kbrock at least outside of miq core team, from my experience a lot of devs use evm:start.
why? I guess because that's what the developer guide says :)
and also how would you test otherwise an entire flow with refresh/events and etc. when developing a provider?

blomquisg:
:+1: yeah, definitely like it better just removed.

Can I get a :+1: from the others on this thread.. @chrisarcand, @Fryguy, @kbrock

jrafanie:
Sorry, I'm late to this.  I'm 👍 removing this line. I don't want to explain how to deal with rails server running while also having a UI worker running.

cben:
Yeah, I can attest that when I first tried to run the project this double instruction confused me.

I'm even more confused now to discover that `rails s` after `evm:start` is not an error — they *both* listen on 3000, the `rails s` appears (?) to get all requests, and after I abort `rail s` 3000 still works thanks to the evm web worker.  What magic is that?!

How about mentioning `rails s` + `rails c` with `simulate_queue_worker` as alternative to evm?
I'm not sure how exactly simulate compares to full evm or minimal mode, but it seems to Just Work.
And it's very simple for a newbie to start debugging — no need to know anything about which worker does what (the [minimal mode page][1] looks scary in that respect), everything that happens visible on stdouts, and byebug works everywhere.

[1]: https://github.com/ManageIQ/guides/blob/master/developer_setup/minimal_mode.md

blomquisg:
Cool, I'm merging this.  We can add additional newbie instructions like @cben refers to later.

Thanks @abonas!

jrafanie:
@cben, I agree.  I feel like we can test drive 80%+ of ManageIQ with either:
* rails console alone
* rails server + rails console (simulate_queue_worker)
* rake evm:start alone

If you're doing smart state analysis or stuff on specific providers or storage locations, these 3 options don't really work as well but for most things, they suffice.  If you want to take a shot and open a PR, I think it's worth mentioning these 3 options in a different section of the guide.

chrisarcand:
+1

I should have said "good idea, but a different section" in my previous
comment :)

On Monday, May 16, 2016, Joe Rafaniello <notifications@github.com> wrote:

> @cben <https://github.com/cben>, I agree. I feel like we can test drive
> 80%+ of ManageIQ with either:
>
>    - rails console alone
>    - rails server + rails console (simulate_queue_worker)
>    - rake evm:start alone
>
> If you're doing smart state analysis or stuff on specific providers or
> storage locations, these 3 options don't really work as well but for most
> things, they suffice. If you want to take a shot and open a PR, I think
> it's worth mentioning these 3 options in a different section of the guide.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/ManageIQ/guides/pull/126#issuecomment-219580369>
>


-- 
Chris Arcand
@chrisarcand

PR #127, opened 2016-06-08
turning StringLiterals on

chriskacerguis:


isimluk:
What are the effects of this change?

isimluk:
Ah, I see. One of the things it does is, it enforces single quotes where double quotes are not needed. That has minor performance effects as these strings do not need to be interpolated.

chriskacerguis:
I spoke to @frayguy going to close this.

PR #130, opened 2016-06-23
Updated packages in Developer Setup

romanblanco:
- npm:

In Fedora 24, *npm* package is already packed in nodejs:
https://bugzilla.redhat.com/show_bug.cgi?id=1339284#c5

- gcc-c++:

I belive we are not using event-machine Gem anymore

correct me if I'm wrong :blush:

### DIFF
--- a/developer_setup.md
+++ b/developer_setup.md
@@ -19,9 +19,9 @@
   sudo dnf -y install postgresql-devel postgresql-server # PostgreSQL Database server and to build 'pg' Gem
   sudo dnf -y install bzip2 libffi-devel readline-devel  # For rbenv install 2.2.0 (might not be needed with other Ruby setups)
   sudo dnf -y install libxml2-devel libxslt-devel patch  # For Nokogiri Gem
-  sudo dnf -y install gcc-c++                            # For event-machine Gem
   sudo dnf -y install sqlite-devel                       # For sqlite3 Gem
-  sudo dnf -y install nodejs npm                         # For ExecJS Gem and bower
+  sudo dnf -y install nodejs                             # For ExecJS Gem and bower
+  rpm -q --whatprovides npm || sudo dnf install npm      # For CentOS 7, Fedora 23 and older
   sudo dnf -y install openssl-devel                      # For rubygems
   sudo dnf -y install cmake                              # For rugged Gem
   ```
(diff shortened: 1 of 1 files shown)


developer_setup.md
```diff
@@ -19,9 +19,8 @@
   sudo dnf -y install postgresql-devel postgresql-server # PostgreSQL Database server and to build 'pg' Gem
   sudo dnf -y install bzip2 libffi-devel readline-devel  # For rbenv install 2.2.0 (might not be needed with other Ruby setups)
   sudo dnf -y install libxml2-devel libxslt-devel patch  # For Nokogiri Gem
-  sudo dnf -y install gcc-c++                            # For event-machine Gem
   sudo dnf -y install sqlite-devel                       # For sqlite3 Gem
-  sudo dnf -y install nodejs npm                         # For ExecJS Gem and bower
+  sudo dnf -y install nodejs                             # For ExecJS Gem and bower
```
romanblanco: this applies only on F24... Should I mention the `npm` package for older distros in comment?

isimluk: The new `nodejs` contains line
```
Obsoletes: npm < 0:3.5.4-6
Provides: npm = %{epoch}:%{npm_version
```
so the `dnf install npm` should continue to work anyway.

While I applaud to the clean-up, I am concerned that the missing npm may confuse centos7 users. What do you think?

romanblanco: @isimluk I've had [issues](https://bugzilla.redhat.com/show_bug.cgi?id=1339284#c0), when I tried installing both `npm` and `nodejs`on my fresh Fedora 24 installation, that's why I'm removing it in the PR.

I agree that the change could be confusing for users with older Fedora or CentOS, I just didn't find the right way to describe it.

Fryguy: @romanblanco So, what do you want to do here?

isimluk: meh, apparently the dnf behaves differently to yum in some [cases](https://bugzilla.redhat.com/show_bug.cgi?id=1096506). Sweet! :candy:

That is because epoch and architecture are not handled consistently. (nodejs is arch, npm is noarch, then there are virtual provides and virtual requires in arch and noarch packages, etc). See [fedora mailing-list](https://lists.fedoraproject.org/archives/list/nodejs@lists.fedoraproject.org/thread/ZVMIY5LM5HXGWCP7FTZCLTAAEKPVHNMW/) for more info.

At this point (when the fedoras are broken), I can think only of this hairy :neckbeard: solution  :hammer:
```
dnf install nodejs
rpm -q --whatprovides npm || dnf install npm
```

romanblanco:
@isimluk I've used your suggestion. Thanks for help :blush:

developer_setup.md
romanblanco: @isimluk I've used your suggestion. Thanks for help :blush:

PR #132, opened 2016-07-07
allow auto-correct of ExtraSpacing

durandom:
because https://github.com/bbatsov/rubocop/issues/2280 is closed


Fryguy:
Verified that this issue was in rubocop as of v0.35.0, so this is good.

PR #134, opened 2016-07-12
Install PostgreSQL 9.4 with software collections

jhernand:
This patch adds documentation explaining how to install PostgreSQL 9.4
in CentOS 7 using SoftwareCollections.org.


jhernand:
This just what I did in my environment, may be useful for others.

PR #136, opened 2016-07-15
Pluggable providers development setup

Ladas:
Pluggable providers development setup


durandom:
the setup works for @Ladas and me.
Plus it has the nice side effect of not changing the `Gemfile`s under git control.

So :+1: from me

PR #141, opened 2016-08-02
Add libcurl for Fedora, used by curb

pilhuhn:
the `Curb` gem wants to install a native extension, which requires `libcurl`.

cc @lucasponce


jrafanie:
FYI, curb dependency was brought in via `ovirt-engine-sdk` added in https://github.com/ManageIQ/manageiq/pull/9521

jrafanie:
Thanks for updating the guides @pilhuhn 👍

PR #143, opened 2016-08-10
Loosen rubocop requirements for less noisy PRs

bdunne:
The bot uses this as the base config for running rubocop on our repos.  In PRs I feel like I frequently see comments about AbcSize and line length.  While some people look at this and make corrections, others completely ignore the comments.  Also, [AbcSize](http://www.rubydoc.info/gems/rubocop/0.37.2/RuboCop/Cop/Metrics/AbcSize) is rarely addressed.

ABC Size > 15 is not hard to do.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -2,6 +2,7 @@
 # Overrides
 #
 AbcSize:
+  Max: 20
   Severity: refactor
 AlignHash:
   EnforcedHashRocketStyle: table


bdunne:
@jrafanie @chrisarcand @Fryguy Opinions?  I feel like I see this one a lot.

Also, I'm on the fence about bumping line length up to 150

chrisarcand:
Can you update description to say exactly this does? I'm just making an assumption without digging it out myself.

chrisarcand:
Ahh here's a good link: http://c2.com/cgi/wiki?AbcMetric

chrisarcand:
I'm indifferent. Ideally you *don't* want it to be too much more than 15, I would think, but admittedly when we're moving around code it seems to pop up and be ignored too often anyway. Sure?

I'm 👎  on line length, I wish we would pay attention to it more often. Beyond 120 is really annoying.

bdunne:
I agree with not being too much more than 15, so I picked 20.  Maybe we should also bump up the severity?
I'm leaning toward not changing line length, so I didn't include it here.

My concern is that people see too many comments and just ignore it.  If we only point out the stuff that is really bad, maybe it will be addressed?

chrisarcand:
> My concern is that people see too many comments and just ignore it. If we only point out the stuff that is really bad, maybe it will be addressed?

I know this feel :( Part of it is needing to reduce cop strictness (like you're suggesting! The more I think on it the more I'm 👍 ) and part of it is we should stop allowing so many to go through. Biggest issue is just when people are moving huge chunks of code from one place to the other and we don't _want_ them to change things for fear of regressions.

bdunne:
Yeah, I agree with moving (ignoring rubocop) in one PR and changing (fixing rubocop) in a second PR.  Right now it's up to the mergers to enforce the rubocop complaints.

sergio-ocon:
It was confusing for me until I did quite a lot of research. And it is possible that the same is happening for external parties.The advice I saw is that you need to refactor your code to call functions instead of having if or switch statements inside the main code.

Perhaps we can add examples into the developers guide and make the bot point to examples where is corrected?

chrisarcand:
> you need to refactor your code to call functions instead of having if or switch statements inside the main code.

And a private interface, exposing only what you originally intended, as well. We have a big issue in this codebase where public class methods are just written wherever you want to 'reduce complexity' only to make spaghetti code later. </rant>

chrisarcand:
ANYWAY, super sorry to hijack this PR for such a lengthy semi-related talk. I'm 👍  on this one, but 🙇  @jrafanie @Fryguy

Fryguy:
I'm considering dropping all of the Metrics cops in general (AbcSize, LineLimit, ClassSize, etc), but this is fine by me for now.

PR #155, opened 2016-09-28
Ran into a missing dependency when setting up on Fedora 24

bdunne:


bdunne:
@carbonin Please review

PR #167, opened 2016-10-19
No postgres user password is required if sudo is available

skateman:
This might return a `could not change directory to` error, but that doesn't interfere with the `psql` command.

### DIFF
--- a/developer_setup.md
+++ b/developer_setup.md
@@ -49,15 +49,13 @@
       to install it from the PostgreSQL Global Development Group repositories.
 
   ```bash
-  sudo passwd postgres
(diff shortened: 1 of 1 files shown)


kbrock:
is there a reason we are running this as postgres.

Was pretty the usermap file allowed root to login as either root or postgres.

skateman:
@kbrock because if you want to run it as root, you need a database named root or an extra `-d postgres` parameter for the `psql`

carbonin:
Plus this is the developer setup, so I don't think we have them set up the user map like we have on the appliance.

PR #172, opened 2016-11-15
Improved i18n documentation readability

romanblanco:
Markdown on GitHub seems to parse differently than parser that is used for ManageIQ.org page, see http://manageiq.org/docs/guides/i18n

There has to be spaces around code blocks

@mzazrivec please review


romanblanco:
@miq-bot assign @mzazrivec

PR #176, opened 2016-12-22
UI split - add PR migration instructions

himdel:
This adds info on how to automatically convert a ManageIQ/manageiq PR to a ManageIQ/manageiq-ui-classic PR. 

This does not preserve comments, labels, assignees, etc. but it *does* preserve the PR description and title, and the commits :).

---

Now also in script form: https://github.com/himdel/dotfiles/blob/master/bin/move-pr


himdel:
Cc @Fryguy

PR #179, opened 2017-01-06
Empty methods don't have to be onelined.

jrafanie:
There are times that onelining a method is fine.  There are times
when it's not.  Let the developer decide when it's appropriate.

cc @bdunne @Fryguy @chrisarcand 

This PR will hopefully cut down on some of the noise seen in many files:

image

### DIFF
--- a/.rubocop_base.yml
+++ b/.rubocop_base.yml
@@ -47,6 +47,8 @@ Style/CollectionMethods:
     reduce: inject
 Style/Documentation:
   Enabled: false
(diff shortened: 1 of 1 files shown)


bdunne:
:+1:

chrisarcand:
👍  (would merge if I could)

isimluk:
👍 (so would I)

chrisarcand:
:shipit:

PR #180, opened 2017-01-11
change docs to plugin development

durandom:
* change notion to plugin development rather than provider development
* make use of `override_gem` method

I also added tweaked it to clone the plugins in a subdirectory ***inside*** your manageiq checkout.
I know it's controversial and debatable. But I think it reduces the chance of breaking other plugins or the core, because doing `ag -r` or `grep -r` or whatever to grep through your directory will also hit the plugins. (Some commands, like `git grep` will not, but alas)

I tend to make that change also to the `developer_setup/classic_ui_split.md` file.

wdyt @martinpovolny @himdel @Fryguy @bdunne 

https://github.com/ManageIQ/manageiq/pull/13193 would add `plugins/` to `.gitignore`

### DIFF
--- a/developer_setup/classic_ui_split.md
+++ b/developer_setup/classic_ui_split.md
@@ -16,18 +16,10 @@ But if you do..
 1. `git remote add upstream git@github.com:ManageIQ/manageiq-ui-classic.git`
 1. `ln -s ../../manageiq spec/`
 1. `cd ../manageiq`
-1. `echo 'unless dependencies.detect { |d| d.name == "manageiq-ui-classic" }' >> Gemfile.dev.rb`
-1. `echo '  gem "manageiq-ui-classic", :path => "'$(cd ../manageiq-ui-classic/; /bin/pwd)'"' >> Gemfile.dev.rb` (because you really need an absolute path there..)
-1. `echo 'end' >> Gemfile.dev.rb`
+1. `echo "override_gem 'manageiq-ui-classic', :path => File.expand_path('../manageiq-ui-classic', __dir__) >> Gemfile.dev.rb`
 1. `bin/update`
 
 And you should be set :).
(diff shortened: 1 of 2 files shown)


developer_setup/classic_ui_split.md
```diff
@@ -16,19 +16,11 @@ But if you do..
 1. `git remote add upstream git@github.com:ManageIQ/manageiq-ui-classic.git`
 1. `ln -s ../../manageiq spec/`
 1. `cd ../manageiq`
-1. `echo 'unless dependencies.detect { |d| d.name == "manageiq-ui-classic" }' >> Gemfile.dev.rb`
-1. `echo '  gem "manageiq-ui-classic", :path => "'$(cd ../manageiq-ui-classic/; /bin/pwd)'"' >> Gemfile.dev.rb` (because you really need an absolute path there..)
-1. `echo 'end' >> Gemfile.dev.rb`
+1. `echo "override_gem 'manageiq-ui-classic', :path => File.expand_path('../manageiq-providers-amazon', __dir__) >> Gemfile.dev.rb`
```
himdel: Well, probably not `../manageiq-providers-amazon` ;)

(And this needs testing, I'm not that sure that the absolute path is no longer needed..)

bdunne: https://github.com/ManageIQ/manageiq/pull/13418 Introducing `override_gem` hasn't been merged yet.

himdel: OK.. but this still points to `manageiq-providers-amazon`..

@durandom ^

durandom: yeah, good catch, thanks...

durandom:
so, @Fryguy hit the button?

PR #181, opened 2017-01-12
UI split - mention an alternative to override_gem

himdel:
Because https://github.com/ManageIQ/guides/pull/180 got merged but https://github.com/ManageIQ/manageiq/pull/13418 didn't yet, so now we have intstructions that can't be used.

Also people may occasionally want to work with an older version.


himdel:
@durandom , @martinpovolny

Fryguy:
override_gem is merged, so I'm closing this

PR #182, opened 2017-01-12
Fixes typo on split directions

d-m-u:
`echo "override_gem 'manageiq-ui-classic', :path => File.expand_path('../manageiq-ui-classic', __dir__) >> Gemfile.dev.rb`

... was missing a double quote.


durandom:
Thanks @d-m-u   🙇  my bad :(

PR #186, opened 2017-01-24
Disable metrics from rubocop.

Fryguy:
We already enable the metrics that we care about via the
.rubocop_cc_base.yml .

@jrafanie Please review


jrafanie:
LGTM, less noise is good, can't merge but I'm 👍

PR #187, opened 2017-01-24
Clarify Ruby style guide reference

Fryguy:
The old text referenced a very old fork of the Ruby style guide, however
it is very difficult to keep that up to date.  Instead, I think it's
easier to reference the current Ruby style guide, and then point out the
limited number of changes we enforce through rubocop.

@jrafanie Please review.

This will also allow me to delete the current fork of the ruby style guide we
have, as I feel it's counterproductive.


chrisarcand:
👍

PR #188, opened 2017-01-26
Fix typo - wrong relative path

chessbyte:


developer_setup/classic_ui_split.md
```diff
@@ -14,7 +14,7 @@ But if you do..
 1. `git clone git@github.com:YOU/manageiq-ui-classic`
 1. `cd manageiq-ui-classic`
 1. `git remote add upstream git@github.com:ManageIQ/manageiq-ui-classic.git`
-1. `ln -s ../../manageiq spec/`
+1. `ln -s ../manageiq spec/`
```
isimluk: I think it is correct and it needs to be this way.

Look at this excerpt from `man ln`.

Symbolic links can hold arbitrary text; if later resolved, a relative link is interpreted in relation to its parent directory.

In this case it is relative to `manageiq/spec/` dir not the current dir of `manageiq/`

Fryguy: @isimluk I'm not sure what you are saying.  @chessbyte actually tried the directions as listed and they don't work.  The only thing that worked was his change above.

jzigmund: @Fryguy @chessbyte it's weird .. but this is not working for me .. I need to use ```ln -s ../../manageiq spec/``` to create symlink with proper path .. maybe some difference between OSX and Fedora in ```ln``` ?

isimluk: I needed to do `ln -s ../../manageiq spec/` as well.

I have no idea from where the differences come from.

isimluk: Man page on MacOS is different. Therefore you don't see the excerpt from `man ln` that I mentioned previously.

himdel:
..unless mac `ln` does something different from linux `ln`

PR #190, opened 2017-02-13
Add hawkular's provider to guides

karelhala:
As I promised @martinpovolny in https://github.com/ManageIQ/manageiq-ui-classic/pull/361 here is guide on how to run hawkinit with and without answers file.

@Jiri-Kremser, @pilhuhn, @theute what do you think?

### DIFF
--- a/providers/hawkular.md
+++ b/providers/hawkular.md
@@ -5,12 +5,12 @@ It's interactive command line tool which uses docker.
 
 For Requirements follow [Hawkinit requirements](https://github.com/Jiri-Kremser/hawkinit#requirements).
 
-### Running Hawkular with Hawkinit tool
+### 1) Running Hawkular with Hawkinit tool
 Install globally trough NPM [Hawkinit](https://www.npmjs.com/package/hawkinit)
 
 ```npm install hawkinit -g```
 
-There are two options how to run hawkinit:
+There are two options how to run **Hawkinit**:
  1. Answer questions from tool
   * Version of hawkular
   * Version of Cassandra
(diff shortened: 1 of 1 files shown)


martinpovolny:
@himdel: please, review

Jiri-Kremser:
looks good. Thanks for it. Perhaps it would worth adding that the h-services are started with a pre-created user with credentials `jdoe:password`. I will add that as well in the hawkinit (it can print that into console).

mtho11:
The only concern is that this is not a supported Hawkular project. 
Should this be part of the Hawkular portfolio before it's in a guide? [I use it myself and think it's great]

karelhala:
@Jiri-Kremser good point! Updated the guide.

@mtho11 That's true, we should wait what @theute and @pilhuhn think about this and we can add some kind of disclosure that this tool is not fully supported and it's just for easy development setup.

providers/hawkular.md
```diff
@@ -73,6 +73,24 @@ There are two options how to run hawkinit:
 }
 
  ```
+
+### Accessing Hawkular
+After **Hawkinit** tool will download, install and run docker files, wait couple of minutes before all services are up
```
Jiri-Kremser: `s/docker files/docker images/`

pilhuhn:
I think Hawkinit is cool - personally I like a docker(-compose) based setup much more, as I can easily dispose it afterwards and the chance of port-conflicts are smaller. 

And we can certainly move that tool either as project to github.com/hawkular or as sub-project inside of github.com/hawkular/hawkular-services

Jiri-Kremser:
> I like a docker(-compose) based setup much more, as I can easily dispose it afterwards and the chance of port-conflicts are smaller.

same here, docker-compose is super easy compared to docker-link-hell or openshift. Once the openshift template is done, I can add the support for it into hawkinit. You know, at the beginning user can select either docker-compose or openshift (+version) as the container orchestration framework, then the other questions can theoretically remain the same.

himdel:
Ok, seems like my previous problem of not being able to access localhost:8080 is fixed now ..
Verified following the instructions does give me a provider I can add to manageiq :+1: .

Although as this is guides, and we probably can't expect all the devs to know *much* about docker, maybe you could add instructions on a) how to stop it (I assume that's just `docker stop $(docker ps | grep -e 'hawk\|cass' | awk '{ print $1 }')`, and b) how to start it again after it's been stoppped (or after reboot) (no idea there).

Jiri-Kremser:
> a) how to stop it 

(SIGINT) Ctrl+C should work: pressed 1 time it tries to shutdown everything grafefuly, pressing it twice it's less merciful

> b) how to start it again after it's been stoppped

The idea right now is that it's stateless and once you stop it, you loose everything (because it's very cheap to spin new containers). But if you want to threat your h-provider as a "pet", you can. If you run the hawkinit, it says something like:

```bash
Later, you can find your hawkular-services listening on http://localhost:8080
Running 'docker-compose up --force-recreate' in directory: /tmp/tmp-11573k3ujXFLACh9z
```

If you navigate to `/tmp/tmp-11573k3ujXFLACh9z`, you can run `docker-dompose up` to start it again. This is not a standard use-case though. (or any other docker-compose command, like `docker-compose logs -f myCassandra`, `docker-compose exec hawkular /bin/bash`, etc.).

Btw. all the Cassandra images has the volume mounted from the host system.

```bash
volumes:
- /tmp/opt/data:/opt/data:rw
```

So the data (hawkular inventory, alerts and metrics) should survive even though you remove the previous containers and start a new ones. If you want to also clean the previous data, do `rm -rf /tmp/opt/data/*`

himdel:
@Jiri-Kremser aaah, thanks, makes sense :) I'm ok with treating them as cattle, as long as we know that's the preferred way :).

@karelhala would you add something like that ^^^ to the PR please? :)

himdel:
@martinpovolny LGTM, if you have the green button...

(Unless you also want `setenforce` mentioned...)

martinpovolny:
Should we link from somewhere to the new article?

himdel:
Oh, good point, from README?

martinpovolny:
probably

himdel:
definitely :) https://github.com/ManageIQ/guides/pull/192

PR #195, opened 2017-03-05
clarify openscap is only needed for some container SSA functionality

cben:
Followup to #189

Without the lib, manageiq will not crash, but some of the container SSA
functionality will be skipped.
AFAICT it affects:
- collection of `OpenscapResult` (`collect_compliance_data`)
- its parsing into `OpenscapRuleResult`s (`create_rule_results`).
- display of these results.


simon3z:
@Fryguy you requested this in #189

@miq-bot assign Fryguy

PR #196, opened 2017-03-15
Added Developer setup guide for Ubuntu / Debian

romanblanco:
@himdel, could you check, if it's correct?

(original source: https://gist.github.com/romanblanco/00df8d0feddb0a7c1b2a8db88d45f61a)

### DIFF
--- a/developer_setup.md
+++ b/developer_setup.md
@@ -11,7 +11,7 @@
   sudo yum -y install dnf
   ```
 
-* Install packages
+* Install Packages
 
   ```bash
   sudo dnf -y install git-all                            # Git and components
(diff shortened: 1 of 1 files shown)


developer_setup.md
```diff
@@ -59,6 +59,52 @@
...
+  sudo apt-get install memcached                        # Memcached for the session store
+  sudo apt-get install postgresql libpq-dev             # PostgreSQL Database server and to build 'pg' Gem
+  sudo apt-get install bzip2 libffi-dev libreadline-dev # For rbenv install 2.2.0 (might not be needed with other Ruby setups)
+  sudo apt-get install libxml2-dev libxslt-dev patch    # For Nokogiri Gem
+  sudo apt-get install libsqlite-dev libsqlite3-dev     # For sqlite3 Gem
+  sudo apt-get install nodejs nodejs-legacy npm         # For ExecJS Gem and bower
+  sudo apt-get install g++                              # For unf Gem
+  sudo apt-get install libcurl4-gnutls-dev              # For Curb
+  sudo apt-get install cmake                            # For rugged Gem
+  sudo apt-get install libgit2-dev pkg-config
```
himdel: Also include `libtool`, for people who don't have it by default.

himdel: + may want to use `apt install` instead of `apt-get install` .. unless we want to support debian stable :)

himdel:
I have not tried installing manageiq so there may still be some packages missing, but .. looks good, and it did work two months ago with these.

PR #199, opened 2017-03-20
[developer_setup] Added steps to install Ruby & Bundler

psachin:


isimluk:
My personal stance on this is that I would prefer not to have these instructions included. They are specific to rbenv or rvm and they may change in future. Further, these instructions are platform specific, MaxOS user may want to choose different path and go with homebrew (whatever that is).

Perhaps we can highlight what is needed. Perhaps we can elaborate what is this all about. When I first saw ruby, I had only vague idea how ruby, rbenv, rvm, gem and bundler play together. Perhaps we can include a paragraph describing what is what and why it is needed, or where to find instructions. I would try to avoid publishing these specific commands.

psachin:
@isimluk 

I totally agree with your thoughts on this. My intention is to help new developers here so that they can setup the environment without browsing other sites. This can work as hint to them. Say

On CentOS/Fedora, Ruby & bundler _can_ be installed using following steps:
```
git clone https://github.com/rbenv/rbenv.git ~/.rbenv
pushd ~/.rbenv
src/configure
make -C src
popd

echo 'export PATH="$HOME/.rbenv/bin:$PATH"' >> ~/.bash_profile
echo 'eval "$(rbenv init -)"' >> ~/.bash_profile
...
```

This will save time. I have no idea about the steps involved on MacOS but I'm sure we can add those steps too. 

I personally think it is better to have steps than have nothing which may discourage newbies to get involve in the project. I certainly do not want this to be a newbie guide :)

Fryguy:
We had a similar discussion in #112, and my stance is the same as @isimluk's.  That is, I don't think we should give explicit instructions, but we may want to highlight or point to installations of ruby installation managers.  The current entry already does a little of this, but could be expanded to point to the actual installation instructions on those sites.

psachin:
@Fryguy @isimluk I totally missed PR #112 . I Should agree with you guys that instead of providing actual steps we can point to appropriate links.

PR #201, opened 2017-03-29
add doc for debugging

ilackarms:
adds instructions on setting up remote/interprocess debugging via `pry-remote`

### DIFF
--- a/developer_setup/debugging.md
+++ b/developer_setup/debugging.md
@@ -1,0 +1,54 @@
+## Debugging ManageIQ with Remote Pry
+
+Interactive debugging using pry or byebug is normally difficult with ManageIQ, as `rake evm:start` launches the server and worker processes in the background. Therefore the only way to do interactive debugging is with `pry-remote` or `byebug` running in remote mode.
+
+Running ManageIQ with `byebug` was exceedingly slow, so this guide will focus on remote debugging using `pry`.
+
+### Adding dependencies
+
+You'll need to add the following gems to your dev overrides file in `bundler.d/`:
+
+```ruby
+  gem 'pry'
+  gem 'pry-remote'
+  gem 'pry-nav'
+```
+
+### Starting the debugger
+
+Setting a breakpoint with `pry` requires adding the following line somewhere in your code:
+
+```ruby
+#code executes until this line
+binding.remote_pry
+#code here does not execute
+```
+
+When execution reaches the `remote_pry` call, the line  `pry-remote] Waiting for client on drb://localhost:9876` will print to STDOUT (stderr?) and execution will block until you connect the client debug process.
+
+To connect the debugger, simply run from the ManageIQ root dir:
+```
+bundle exec pry-remote
+```
+which will attempt to connect to the pry server on port 9876.
+
+With the addition of `pry-nav`, you'll have access to commands `continue`, `step`, and `next`, to navigate through your code.
+
+### Notes
+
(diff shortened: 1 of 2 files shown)


ilackarms:
@Fryguy @cben please review my submission. Thank you

cben:
What if another worker hits a breakpoint too? Can pry find an available port?

ilackarms:
good question. afaik pry always uses the same port `9876`, so the 2nd worker will panic if it hits a breakpoint while you're the first bp is still being debugged. conditional breakpoints are a possible workaround for this.

developer_setup/debugging.md
```diff
@@ -0,0 +1,64 @@
+## Debugging ManageIQ with Remote Pry
+
+Interactive debugging using pry or byebug is normally difficult with ManageIQ, as `rake evm:start` launches the server and worker processes in the background. Therefore the only way to do interactive debugging is with `pry-remote` or `byebug` running in remote mode.
+
+Running ManageIQ with `byebug` was exceedingly slow, so this guide will focus on remote debugging using `pry`.
+
+### Adding dependencies
+
+You'll need to add the following gems to your `Gemfile`:
```
Fryguy: Instead of requesting the user to modify the Gemfile, prefer creating/modifying `Gemfile.dev.rb` or create a file in the `bundler.d` directory

developer_setup/debugging.md
```diff
@@ -0,0 +1,64 @@
...
+    gem 'pry-nav'
+    # ...
+  end
+```
+
+The `pry-remote` gem must be required at some point before your (first) breakpoint. A convenient place to put this is in `config/environments/development.rb`:
+
+```ruby
+require 'pry-remote'
+```
```
Fryguy: Is this part necessary?  When you do `gem 'pry-remote'`, then require 'pry-remote' is done automatically, unless you do `:require => false`.

Fryguy:
This is pretty awesome @ilackarms !

developer_setup/debugging.md
gberginc: I second adding this to `bundler.d` dir. It's also in line with the [recent change](https://github.com/ManageIQ/manageiq/commit/3a19899cef6977eeaf737517ba4b80750d283bcb).

durandom:
@ilackarms you want to address the small changes so we can get this published? That would be awesome 😄

developer_setup/debugging.md
ilackarms: for some reason, when i tried without `require`ing, i got `undefined local variable` errors. maybe I was doing something wrong?

durandom:
@chrisarcand @Fryguy can we merge this? 
I'd like to add some info on debugging via foreman / `ruby lib/workers/bin/run_single_worker.rb `

developer_setup/debugging.md
chrisarcand: Yeah something is fishy here. You shouldn't need to re-require this. Double check?

ilackarms: tried it again and no longer had this issue. not sure how i encountered it the first time. anyway, removed from readme :)

ilackarms:
@chrisarcand I've tried it with pry-byebug. There's actually a bug with pry-byebug that causes remote-pry to immediately exit without bringing up a repl. using pry-nav fixes the issue

chrisarcand:
@ilackarms Huh. Ok! Thanks for checking. 👍

PR #202, opened 2017-04-06
Plugin development

durandom:
Updated guides on plugin development
Removed guide on ui split, as it's handled by plugin development guide

@miq-bot assign @Fryguy


developer_setup/plugins.md
```diff
@@ -0,0 +1,154 @@
...
+lets you separate checkouts cleanly.
+
+```bash
+cd ~/src
+git clone git@github.com:JoeSmith/manageiq-providers-amazon.git
+git clone git@github.com:JoeSmith/manageiq-content.git
+git clone git@github.com:JoeSmith/manageiq-ui-classic.git
+```
+
+#### Inside the core manageiq checkout
```
durandom: If you want, I can remove this part altogether.

developer_setup/plugins.md
```diff
@@ -0,0 +1,154 @@
...
+```bash
+cd ~/src/manageiq ; mkdir plugins
+git clone git@github.com:JoeSmith/manageiq-providers-amazon.git plugins/manageiq-providers-amazon
+git clone git@github.com:JoeSmith/manageiq-content.git plugins/manageiq-content
+git clone git@github.com:JoeSmith/manageiq-ui-classic.git plugins/manageiq-ui-classic
+```
+
+### Running tests
+
+To run the tests for a plugin, we need an application context. Usually rails expects a dummy rails app. But because
```
Fryguy: When referring to rails, prefer capitalizing: Rails

developer_setup/plugins.md
```diff
@@ -0,0 +1,154 @@
...
+our plugins are only for manageiq, we run the tests inside the manageiq app. Therefore we need a checkout of manageiq
```
Fryguy: When referring to manageiq as a project, prefer `ManageIQ`...when referring to the directory, prefer `manageiq`

developer_setup/plugins.md
```diff
@@ -0,0 +1,154 @@
...
+bin/update
+# == Updating manageiq sample app ==
+```
+
+### Cross repository dependencies
+
+Sometimes you need to develop your feature against a branch of manageiq that is not yet merged into master. This creates
+cross repository dependencies, which can be handled from the plugin side or from the core side.
+
+#### Symlink you sample app
```
Fryguy: Typo `you` => `your`

developer_setup/plugins.md
```diff
@@ -0,0 +1,154 @@
...
+
+Creating a symlink from `spec/manageiq` to your current checkout of manageiq will run the tests inside whatever branch
+you checked out in the core repo.
+
+```bash
+# remove current sample app
+rm -rf spec/manageiq 
+# create a symlink
+ln -s ~/src/manageiq spec/manageiq
```
Fryguy: Wherever we have `~/src/manageiq`, let's prefer `/path/to/manageiq`.  Then there is no confusion.  Additionally, down below on line 91 you mentuion override gem, which actually won't work with "~/src" as written

developer_setup/plugins.md
Fryguy: It's fine :+1:

developer_setup/plugins.md
Fryguy: This applies to all of the mentions of rails and manageiq elsewhere in the doc as well.

developer_setup/plugins.md
```diff
@@ -0,0 +1,154 @@
...
+```
+
+#### Dependency on a local gem
+
+Inside manageiq core, you can override gem dependencies with `override_gem`. This is a small helper to be used inside
+`Gemfile.dev.rb`. Use it to point the dependency to a local version of your plugin. 
+
+```ruby
+# Gemfile.dev.rb:
+override_gem 'manageiq-providers-amazon', :path => '~/src/manageiq-providers-amazon'
```
durandom: ok, changing to generic path. But actually that works, I have this in my Gemfile.dev.rb - except for double quotes. Does that influence tilde expansion?

durandom:
@Fryguy thanks for the review. Addressed your comments in the last commit

developer_setup/plugins.md
```diff
@@ -0,0 +1,153 @@
...
+```
+
+#### Dependency on a local gem
+
+Inside ManageIQ core, you can override gem dependencies with `override_gem`. This is a small helper to be used inside
+`Gemfile.dev.rb`. Use it to point the dependency to a local version of your plugin. 
+
+```ruby
+# Gemfile.dev.rb:
+override_gem 'manageiq-providers-amazon', :path => File.expand_path("/path/to/manageiq-providers-amazon")
```
Fryguy: Nice.  Even better.

PR #204, opened 2017-04-07
Adds minishift instructions to the openshift developer guides

NickLaMuro:
Also fixes some typos and cleans up the original docs for openshift.

~~Note:  There is currently a [bug with minishift](https://github.com/minishift/minishift/issues/710) when using the described addon.  Eventually, the idea is that this guide can be trimmed down if this addon is included in the core `minishift` project, but for now, a workaround has been included to make this work regardless if that bug is fixed or not.~~

This bug has been isolated [here](https://github.com/minishift/minishift/issues/729), and the current workaround is to just not used string based args with spaces (which I have updated the addon to do).

### DIFF
--- a/providers/openshift.md
+++ b/providers/openshift.md
@@ -49,3 +49,84 @@ inventory/byo/hosts.example.
 ```console
 $ ansible-playbook playbooks/byo/config.yml -i path/to/inventory/file
 ```
+
+
+### Running with minishift
+
+
+For a simple single host setup for working with the openshift provider locally,
+you can use [`minishift`](https://github.com/minishift/minishift) to run a
+local openshift instance.
+
+
+#### Requirements
+
+* [`minishift`](https://github.com/minishift/minishift) >= 1.0.0.rc1 ([Installation Instructions](https://docs.openshift.org/latest/minishift/getting-started/installing.html#installing-instructions))
+* Virtualization host of your choice (see supported hypervisors [here](https://docs.openshift.org/latest/minishift/getting-started/installing.html#install-prerequisites))
+
+
+#### Quickstart
+
+Since VirtualBox is a consistent hypervisor across all platforms, this guide
+will assume that is being used, and updates to some commands might be
+necessary (most commands should be virtualization software agnostic and work
+regardless of the hypervisor).
+
+1. Download the manageiq.addon for minishift to a directory of your choosing:
+  
+  ```console
+  
+  $ git clone https://gist.github.com/e2fac8be87ea0e9f429b6f5d75e02176.git path/to/minishift-addons/manageiq
+  ```
+
+  You can inspect the addon here:
+  
+    https://gist.github.com/NickLaMuro/e2fac8be87ea0e9f429b6f5d75e02176
+  
+  It is recommended to keep a directory of your addons in one place.
+  
+2. Add and enable the manageiq addon to minishift:
+  
+  ```console
+  $ minishift addons install --force path/to/minishift-addons/manageiq
+  $ minishift addons enable manageiq
+  ```
+  
+3. Start minishift:
+  
+  ```console
+  $ minishift start --vm-driver virtualbox --openshift-version "v1.5.0-rc.0"
+  ```
+  
+4. Grab the minishift IP:
+  
+  ```console
+  $ minishift ip
+  ```
+  
+5. Grab the token to access openshift through `manageiq`:
+  
+  ```console
+  $ ~/.minishift/cache/oc/v1.5.0-rc.0/oc login -u system:admin
+  $ ~/.minishift/cache/oc/v1.5.0-rc.0/oc sa get-token -n management-infra management-admin
+  ```
+  
+  Or in a single command form:
+  
+  ```console
+  $ (export PATH="~/.minishift/cache/oc/v1.5.0-rc.0/:$PATH"; oc login -u system:admin > /dev/null; oc sa get-token -n management-infra management-admin)
+  ```
+  
+6. Configure a provider in ManageIQ, filling in your token and IP where
+   appropriate:
+  
+  ```console
+  $ bin/rails c
+  irb> token = '<<YOUR_TOKEN_FROM_ABOVE_HERE>>'
+  irb> host  = '<<YOUR_IP_FROM_ABOVE_HERE>>'
+  irb> os = ManageIQ::Providers::Openshift::ContainerManager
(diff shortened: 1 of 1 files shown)


NickLaMuro:
@moolitayer Since you were the original author on this guide, curious about your thoughts on this one.

NickLaMuro:
> After seeing the instructions, I want to add to some scripts/tools to simplify the instructions

@kbrock Heh, actually, I am way ahead of you on that one.  Been working through some things yesterday in a [minishift issue](https://github.com/minishift/minishift/issues/710) to get this to be a core addon, but might have some other ways that I can reduce the steps in this PR if that doesn't go through.

Also, after doing some more research and isolating the source of the bug mentioned in the description, I can reduce some of the steps a bit since they are a bit excessive.

providers/openshift.md
```diff
@@ -32,9 +41,146 @@ master_hostname
 node_hostname
 ```
 
-openshift_use_manageiq is set to True to configure the Service Account needed by ManageIQ. there are more like variables like use_metrics to enable other options in OpenShift.
-A more detailed example will be found in inventory/byo/hosts.example.
+openshift_use_manageiq is set to True to configure the Service Account needed
+by ManageIQ. there are more like variables like use_metrics to enable other
+options in OpenShift.  A more detailed example will be found in
+inventory/byo/hosts.example.
```
moolitayer: Not really related to your addition but it would be nice to change a couple of things that changed if you have the time.
`openshift_use_manageiq` is no longer needed as it is set by default in openshift-ansible[1]
I think we can update the above paragraph to:

A more detailed example can be found in inventory/byo/hosts.origin.example. Specifically, to deploy metrics and logging use `openshift_hosted_metrics_deploy=True` and `openshift_hosted_logging_deploy=True`.

providers/openshift.md
```diff
@@ -32,9 +41,146 @@ master_hostname
...
+
+* [`minishift`](https://github.com/minishift/minishift) >= 1.0.0.rc1 ([Installation Instructions](https://docs.openshift.org/latest/minishift/getting-started/installing.html#installing-instructions))
+* Virtualization host of your choice (see supported hypervisors [here](https://docs.openshift.org/latest/minishift/getting-started/installing.html#install-prerequisites))
+
+
+#### Quickstart
+
+Since VirtualBox is a consistent hypervisor across all platforms, this guide
+will assume that is being used, and updates to some commands might be
+necessary.  Most commands should be virtualization software agnostic and work
```
moolitayer: double space before `Most`

providers/openshift.md
```diff
@@ -32,9 +41,146 @@ master_hostname
...
+      "name": "management-infra-admin"
+    },
+    "rules": [
+      {"resources": ["pods/proxy"],"verbs":["*"]}
+    ]
+  }
+  ```
+  
+3. Create a hawkular-metrics-admin-cluster-role.json file with the following
+   contents in that dir:
```
moolitayer: s/with the following contents in that dir/with the following content/

providers/openshift.md
```diff
@@ -32,9 +41,146 @@ master_hostname
...
+    },
+    "rules": [
+      {
+        "apiGroups": [""],
+        "resources": ["hawkular-metrics","hawkular-alerts"],
+        "verbs":     ["*"]
+      }
+    ]
+  }
+  ```
```
moolitayer: will the minishift deployment described here contain a metrics component?

NickLaMuro: Probably not, but I was mostly just extracting/reverse engineering exactly what I saw in the Ansible playbook into a script since I probably spent a good day trying to figure out a proper permission configuration to get `manageiq` a token to do a refresh...

Might be nice though to have an addition to this setup (at somepoint) to deploy hawkular instance, and connect it so that developers can mess with that as well.  Probably will do that in a separate PR.

NickLaMuro:
@moolitayer These are all great suggestions, and I will probably implement most of them (plus a few that I have planned on doing anyway), but will respond in details to each inline:

> Too bad I will miss the talk on minishift due to Passover, I hope there would be a recording.

Oh, I was unaware this was a thing.  I will have to look into seeing that myself.

> It might be nice to add in the document top a short section with links explaining the different deployments. From the top of my head (I'm sure you can phrase that better)
> 
> ...
> 
> Optionally we could add a third option (just a link is a good start): openshift's oc cluster up

I like both of these, and will probably do something like that.  Thanks for the suggestion!

providers/openshift.md
NickLaMuro: Yeah, I can do that.  Thanks.

providers/openshift.md
```diff
@@ -32,9 +41,146 @@ master_hostname
...
+necessary.  Most commands should be virtualization software agnostic and work
+regardless of the hypervisor.
 
-openshift_use_manageiq is set to True to configure the Service Account needed by ManageIQ. there are more like variables like use_metrics to enable other options in OpenShift.
-A more detailed example will be found in inventory/byo/hosts.example.
-```ansible-playbook playbooks/byo/config.yml -i path/to/inventory/file```
+1. Create an addon directory for manageiq in minishift.  Something like:
+  
+  ```console
+  $ mkdir -p ./minifshift/addons/manageiq
```
moolitayer: s/minifshift/minishift/  (or replace the instructions referring to this below. caught this when running locally)

providers/openshift.md
```diff
@@ -32,9 +41,146 @@ master_hostname
...
+  
+  ```console
+  $ minishift addons install ./minishift/addons/manageiq
+  $ minishift addons enable manageiq
+  ```
+  
+5. Start minishift
+  
+  ```console
+  $ minishift start --vm-driver virtualbox  --openshift-version "v1.5.0-rc.0"
```
moolitayer: Maybe we are better off not specifying a version? that way when we don't update this page it still works?

NickLaMuro: Only did this for "copy pasta"-able instructions.  If you notice some of the other commands actually use the `oc` command that get's cached by minishift, so that was intentional.  Reduces the dependencies, but causes some weird workarounds.

Open to suggestions for a better way of handling this.

providers/openshift.md
NickLaMuro: derp... yeah, will change

NickLaMuro: This is another reason I am pushing this to be added to the core part of the repo as well, since these instructions are a bit derpy...  https://github.com/minishift/minishift/issues/710

moolitayer:
@NickLaMuro I created https://github.com/ManageIQ/guides/pull/205 for better visibility of this page.

PR #209, opened 2017-05-02
Add help for sys-proctable installation issues

djberg96:
As requested, this adds help for issues installing the `sys-proctable` gem.


juliancheal:
LGTM 👍

PR #212, opened 2017-05-16
Add openshift link to "Provider setup instructions"

himdel:
@NickLaMuro added some openshift setup instructions (thanks!), which means we should be mentioning openshift in the "provider setup instructions" list in README :).


NickLaMuro:
Seems reasonable.  👍

PR #228, opened 2017-07-12
Enable Style/MethodCallWithArgsParentheses with exceptions

bdunne:
Enforce using parentheses for methods that have arguments.  I added a bunch of DSL type methods to the exception list.

### DIFF
--- a/.rubocop_base.yml
+++ b/.rubocop_base.yml
@@ -62,6 +62,101 @@ Style/HashSyntax:
   EnforcedStyle: hash_rockets
 Style/IfUnlessModifier:
   Enabled: false
+Style/MethodCallWithArgsParentheses:
+  Exclude:
+    - Gemfile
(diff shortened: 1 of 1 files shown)


.rubocop_base.yml
```diff
@@ -62,6 +62,101 @@ Style/HashSyntax:
...
+    - serialize
+    - skip_before_action
+    - validate
+    - validates
+    # Gems
+    - acts_as_tree
+    - default_value_for
+    - default_values
+    - has_ancestry
+    # MIQ
```
chrisarcand: So thorough. 😄

bdunne: It was painful to generate the list.  I know it's not complete, but it's a starting point.

jrafanie: wow, crazy.  There's so many dsl methods...

PR #230, opened 2017-07-19
Adds yarn install instructions for Fedora / CentOS

petrblaho:
I think that this was missing.

### DIFF
--- a/developer_setup.md
+++ b/developer_setup.md
@@ -29,10 +29,10 @@
   sudo dnf -y install openscap                           # Optional, for openscap Gem for container SSA
   ```
 
-* Install the _Bower_ package manager
+* Install the _Bower_ and _Yarn_ package manager
 
   ```
-  sudo npm install -g bower
+  sudo npm install -g bower yarn
   ```
 
 * Enable Memcached


himdel:
Thanks, these were previously only mentioned in the Debian/Ubuntu section, should be here as well.

mtho11:
The recommended `yarn` install is natively and not through npm: https://yarnpkg.com/lang/en/docs/install/

petrblaho:
@mtho11 this is interesting... perhaps we need issue for this b/c that covers not only Fedora / CentOS but other systems as well. What do you think, @himdel ?

himdel:
Well.. it definitely makes sense to link to those instructions .. and I notice `yarn` is still missing from the mac section.

OTOH I'd keep the `npm install -g yarn` there as well, simply because these instructions should be complete by themselves, without having to recursively read other instructions, and the npm way is simpler than add a key, add a repo, and install.

petrblaho:
@himdel ok. I will do that.

Is that fine with you, @mtho11 ?

mtho11:
Perfect!

PR #232, opened 2017-07-20
Improve readability by disabling Style/TrailingUnderscoreVariable

bdunne:
When enabled, it will output the following:
```
test.rb:1:7: C: Do not use trailing _s in parallel assignment. Prefer a, b, = 1, 2, 3, 4.
a, b, _, _ = 1, 2, 3, 4
      ^^^^^
```

Thanks for mentioning this one @carbonin

### DIFF
--- a/.rubocop_base.yml
+++ b/.rubocop_base.yml
@@ -214,6 +214,8 @@ Style/TrailingCommaInArguments:
   Enabled: false
 Style/TrailingCommaInLiteral:
   Enabled: false
+Style/TrailingUnderscoreVariable:
+  Enabled: false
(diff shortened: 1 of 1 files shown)


jrafanie:
So, this PR will allow you to use `a, b, _, _ = 1, 2, 3, 4` without the hassle from the bot?  I'm Fine with that.  `a, b, = 1, 2, 3, 4` still looks like a syntax error in my brain so I prefer to have "something" there.

bdunne:
> `a, b, = 1, 2, 3, 4` still looks like a syntax error in my brain so I prefer to have "something" there.

^^^ Exactly 👍

PR #235, opened 2017-07-24
ui - add Forms document

himdel:
This adds a document on the current state of forms, conversion routes between the various stages of angular conversion, and a tentative plan for the Angular (2+) future.


martinpovolny:
ping @himdel : can we merge a part?

himdel:
@martinpovolny Sorry for the delay, dropped the WIP bits, so this should now be ready.

(The rest of the changes will come through https://github.com/ManageIQ/guides/pull/267.)

PR #240, opened 2017-08-04
Updated guides with method for fetching a single item with RBAC

romanblanco:
Updating guides after https://github.com/ManageIQ/manageiq-ui-classic/pull/1636


romanblanco:
@himdel or @martinpovolny can you review?

PR #242, opened 2017-08-10
Developer setup - solutions for a bunch of Mac problems

himdel:
I tried installing miq on a mac in school today, this is the list of things that had to be fixed - should probably be mentioned in the guides.

* With ruby installed as a different user, you can't `bundle install` anything - a workaround is to install the bundle locally
* Nokogiri 1.7.2 bundles it's own version of libxml2 on macs, but fails to compile on a missing LZMA_OK constant
* sys-proctable tries to install an aix version by default and fails

### DIFF
--- a/developer_setup.md
+++ b/developer_setup.md
@@ -299,3 +299,32 @@ If this happens, you may be missing developer tools in your OS X. Try to install
 
 If this happens it may be a Bundler issue. Try running `bundle config specific_platform true`
 and re-try.
+
+* Can't install any gems during bundle install
+
+```
+# install all dependencies to a local path
+bundle install --path vendor/bundle
+```
+
+* Can't install `nokogiri-1.7.2` on a Mac
+
+```
+# install dependencies
+brew install pkgconfig
+brew install libxml2
+
+# test the gem can be compiled with the right libxml2
+export PKG_CONFIG_PATH=/usr/local/opt/libxml2/lib/pkgconfig
+gem install nokogiri -v 1.7.2 -- --use-system-libraries
+
+# configure bundle to do so
+bundle config build.nokogiri --use-system-libraries --with-xml2-include=$(brew --prefix libxml2)/include/libxml2
+```
+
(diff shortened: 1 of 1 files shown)


developer_setup.md
```diff
@@ -176,6 +176,37 @@
   launchctl load -w ~/Library/LaunchAgents/homebrew.mxcl.memcached.plist
   ```
 
+##### Problems
```
Fryguy: There is already a Troubleshooting section at the bottom...can you please move this down there?

himdel:
@Fryguy thanks, moved to Troubleshooting (and updated the mac-only problems to say so)

developer_setup.md
```diff
@@ -299,3 +299,32 @@ If this happens, you may be missing developer tools in your OS X. Try to install
 
 If this happens it may be a Bundler issue. Try running `bundle config specific_platform true`
 and re-try.
+
+* Can't install any gems during bundle install
```
chrisarcand: Is there a more specific error message that presents itself in this scenario you outlined in the PR description? This is a fairly broad statement.

himdel: Sorry, I no longer have access to the macs :(.

And trying locally, it seems it's trying to just run sudo for me, which is definitely not what I saw on the mac.

(Also, is it really that broad? `bin/setup` takes care of installing bundler, and we don't need to compile native extensions for most of the gems, so.. I can't actually come up with any other reason for *all* gems to fail installing other than permissions.)

chrisarcand: I think it's broad as there may be many reasons all your gems won't install. [Here's the first hit I got googling](https://stackoverflow.com/questions/2414130/cant-install-any-gems), for example.

If you don't have access and can't get the exact message, that's ok though! Thought I'd ask anyway.

himdel: Sorry, I don't .. but I should be able to re-try in september, so I'll do a follow-up PR then (I hope :)).

PR #245, opened 2017-08-25
Adds note to plugins guide about relative path with override_gem

petrblaho:
Using only relative path without `__dir__` causes issues when gems are
loaded from different place.

### DIFF
--- a/developer_setup/plugins.md
+++ b/developer_setup/plugins.md
@@ -109,9 +109,16 @@ ln -s /path/to/manageiq spec/manageiq
 Inside ManageIQ core, you can override gem dependencies with `override_gem`. This is a small helper to be used under
 `bundler.d` directory. Use it to point the dependency to a local version of your plugin. 
 
+Please, use absolute path in
 ```ruby
 # bundler.d/local_plugins.rb:
-override_gem 'manageiq-providers-amazon', :path => File.expand_path("/path/to/manageiq-providers-amazon")
+override_gem 'manageiq-providers-amazon', :path => File.expand_path('/home/developer/repos/manageiq-providers-amazon')
+```
+
+or use relative path to current file and `__dir__` like
+```ruby
+# bundler.d/local_plugins.rb:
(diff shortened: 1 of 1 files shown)


andyvesel:
Could that be more visually clear if we visually split those two notes about absolute and relative paths like this?

#### Dependency on a local gem
     
Inside ManageIQ core, you can override gem dependencies with `override_gem`. This is a small helper to be used under `bundler.d` directory. Use it to point the dependency to a local version of your plugin. 

Please, use absolute path in 
```ruby
# bundler.d/local_plugins.rb:
override_gem 'manageiq-providers-amazon', :path => File.expand_path('/home/developer/repos/manageiq-providers-amazon')
```
or use relative path to current file and `__dir__` like 
```ruby
# bundler.d/local_plugins.rb:
override_gem 'manageiq-providers-amazon', :path => File.expand_path('../..manageiq-providers-amazon', __dir__')
```

petrblaho:
@andyvesel you are right....

developer_setup/plugins.md
```diff
@@ -109,9 +109,16 @@ ln -s /path/to/manageiq spec/manageiq
...
 ```ruby
 # bundler.d/local_plugins.rb:
-override_gem 'manageiq-providers-amazon', :path => File.expand_path("/path/to/manageiq-providers-amazon")
+override_gem 'manageiq-providers-amazon', :path => File.expand_path('/home/developer/repos/manageiq-providers-amazon')
+```
+
+or use relative path to current file and `__dir__` like
+```ruby
+# bundler.d/local_plugins.rb:
+override_gem 'manageiq-providers-amazon', :path => File.expand_path('../../manageiq-providers-amazon', __dir__')
```
Fryguy: Remove the bonus quote mark after __dir__ that causes a SyntaxError

petrblaho: Thanks @Fryguy , I missed that.

PR #246, opened 2017-09-06
Add git how to part and delete it in TODO

ailisp:
This is mostly from Yuri and part of them from searching. It's tested during my daily practice.


ailisp:
\cc @Fryguy Please review, thanks.

PR #249, opened 2017-09-07
rubocop rule to ignore some rake methods

jameswnl:


jameswnl:
@miq-bot add_labels enhancement

PR #253, opened 2017-10-02
Don't document installation steps of ruby version managers

bdunne:
More accurately specify that the version dependencies are minimums


bdunne:
Yeah, I was going down the list one-by-one

PR #256, opened 2017-10-03
add doc for setting up kubernetes

ilackarms:
Adds a small doc for prerequisites in order to use the ManageIQ Kubernetes Provider


ilackarms:
@simon3z @cben cc if you're interested

providers/kubernetes.md
```diff
@@ -0,0 +1,55 @@
...
+    kubectl create sa -n management-infra management-admin   
+    kubectl create sa -n management-infra inspector-admin   
+    ```
+0. Grant cluster-reader cluster role to `management-admin` SA:
+    ```bash
+    kubectl create clusterrolebinding management-infra-cluster-reader --clusterrole=cluster-reader --user=system:serviceaccount:management-infra:management-admin
+    ```
+0. Retrieve the serviceaccount token for `management-admin` (this will be the auth token ManageIQ uses):
+    ```bash
+    kubectl describe secret -n management-infra $(kubectl get secrets -n management-infra | grep management-admin | cut -f1 -d ' ') | grep -E '^token' | cut -f2 -d':' | tr -d '\t'
```
simon3z: @ilackarms can you check with Kubernetes/OpenShift engineers? I have the feeling that this should be simpler.

simon3z:
Nice!

providers/kubernetes.md
ilackarms: you can use `kubectl describe secret -n management-infra management-admin-token-XXXX` and then just copy the text next to `token` in the output.

this command saves having to look up the secret name, and having to separate the token from the rest of the output. it's useful if the user wants to store the token in a variable (such as for use with `curl`).

moolitayer: Maybe `oc sa get-token -n management-infra management-admin` can help here?

moolitayer: oh this is for bare k8s...

moolitayer:
@ilackarms awesome!
you can add another link in the containers page [1] for more visibility (right below the OpenShift link)
[1] https://github.com/ManageIQ/guides/blob/87ad4dff1a039ea246c5e844dd69251d7e110730/providers/containers.md

PR #203, opened 2017-04-06
Plugin development

durandom:
* added section on remotes
* change remotes to https, because you cannot push to https by default, which makes it safer for people with merge perms on upstream

@Fryguy

PR #260, opened 2017-10-17
Use https remote url when setting up upstream for manageiq-ui-classic

elad661:
**Description**

While setting up my environment, I copied that line and got an error, because ssh remote urls only work for people with commit access.
It's better to use the https url in a guide that is probably mostly intended for people without commit access.

### DIFF
--- a/developer_setup/plugins.md
+++ b/developer_setup/plugins.md
@@ -52,7 +52,7 @@ cd manageiq-ui-classic
 git remote -v
 # origin  git@github.com:JoeSmith/manageiq-ui-classic.git (fetch)
 # origin  git@github.com:JoeSmith/manageiq-ui-classic.git (push)
-git remote add upstream git@github.com:ManageIQ/manageiq-ui-classic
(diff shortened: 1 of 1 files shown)


elad661:
@durandom @simon3z @cben please review.

cben:
LGTM :+1:

I'm not sure it's commit access, I just cloned it via ssh but don't have commit access.  Perhaps it's being in the Org?  Or having generated and uploaded an SSH key to github?

Whatever the reason, we want everyone to be able to clone & github recommends HTTPS.
https://help.github.com/articles/which-remote-url-should-i-use/

durandom:
I wanted `https` as well, but for the reason of not pushing accidentially to upstream.
See https://github.com/ManageIQ/guides/pull/203#pullrequestreview-31430108

And we went for `ssh`. 

To your issue: I have no problem cloning any repos from github via ssh. It must be a problem of your local setup. Have a look here https://help.github.com/articles/testing-your-ssh-connection/

PR #270, opened 2017-11-09
Change `dnf` to `yum` for CentOS users

naltun:
`dnf` is no longer available for CentOS 7, so changed the documentation to tell CentOS users to use `yum` instead.


naltun:
@agrare Ah, right. Let me change that.

naltun:
#271

PR #271, opened 2017-11-09
Add 'use yum' advice notice for CentOS users

naltun:
`dnf` is no longer available for CentOS 7 users. Advise users to use `yum` instead.

### DIFF
--- a/developer_setup.md
+++ b/developer_setup.md
@@ -4,7 +4,7 @@
 
 #### Fedora / CentOS 7
 
-* CentOS only - Enable EPEL & install DNF
+* CentOS only - dnf is no longer available on CentOS7. In the examples below, replace dnf with yum
 
   ```bash
   sudo rpm -ivh https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
--- a/developer_setup.md
+++ b/developer_setup.md
@@ -4,12 +4,7 @@
 
 #### Fedora / CentOS 7
 
-* CentOS only - dnf is no longer available on CentOS7. In the examples below, replace dnf with yum
-
-  ```bash
-  sudo rpm -ivh https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
-  sudo yum -y install dnf
-  ```
+* CentOS only - dnf is no longer available on CentOS 7. In the examples below, replace dnf with yum.
 
 * Install Packages
 
--- a/developer_setup.md
+++ b/developer_setup.md
@@ -4,7 +4,7 @@
 
 #### Fedora / CentOS 7
 
-* CentOS only - dnf is no longer available on CentOS 7. In the examples below, replace dnf with yum.
+* CentOS only - use yum instead of dnf.
 
 * Install Packages
 


naltun:
So, I was trying to work with CentOS 7 by using `dnf` to install all dependencies. After some chatting in Gitter/manageiq, we discovered that `dnf` is no longer available for CentOS. Here's an update to the Development Setup guide to reflect this.

developer_setup.md
```diff
@@ -4,7 +4,7 @@
 
 #### Fedora / CentOS 7
 
-* CentOS only - Enable EPEL & install DNF
+* CentOS only - dnf is no longer available on CentOS7. In the examples below, replace dnf with yum
 
   ```bash
   sudo rpm -ivh https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
```
agrare: Should be able to delete this section instructing to install epel and dnf now

naltun:
Yep, I removed that other section @agrare.

agrare:
@miq-bot assign @Fryguy

developer_setup.md
```diff
@@ -4,12 +4,7 @@
 
 #### Fedora / CentOS 7
 
-* CentOS only - Enable EPEL & install DNF
-
-  ```bash
-  sudo rpm -ivh https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
-  sudo yum -y install dnf
-  ```
+* CentOS only - dnf is no longer available on CentOS 7. In the examples below, replace dnf with yum.
```
chrisarcand: This isn't accurate.

`dnf` is `yum`'s successor; it's that it's *not the default* in CentOS 7. CentOS 7 is directly based off of Red Hat Enterprise Linux 7. RHEL 7 is based off of Fedora 19, and most importantly `dnf` wasn't changed as the default of Fedora until Fedora 23.

TL;DR: Fedora is leading the charge having replaced `yum` with `dnf` and it's all but a matter of time before the next version of RHEL, and hence the next version of CentOS, uses `dnf`.

I'm confused as to why this change is necessary as the stuff you removed below is actually the right way to just use `dnf` on CentOS instead of `yum`.

chrisarcand: I should add that if you do suggest just saying 'use yum' for CentOS 7 and want to remove the EPEL stuff, that's fine, but I'd just say "For CentOS, `yum` is the default package manager; If you're using CentOS, replace dnf with yum in all the examples below"

naltun: Well, earlier (if I recall, you were active in that Gitter chat) myself and a few others tried installing `dnf` on CentOS 7 and a `no package found` error was given. Someone, I believe, pointed out that CentOS removed the `dnf` package. Even with getting the `EPEL` set up, `dnf` wasn't found as a valid package. Therefore, `yum` is the _only_ option, no?

agrare: > I'm confused as to why this change is necessary as the stuff you removed below is actually the right way to just use dnf on CentOS instead of yum. Are you saying they literally removed the availability of the rpm yum? That doesn't seem right.

They removed the RPM from EPEL 7 because it was old and buggy, it was never available in CentOS or RHEL core.  The "right" way to install packages in CentOS 7 is yum.

I suspect that we just suggested installing `dnf` so that we didn't need to have a section for CentOS and a section for Fedora where the only difference was which package manager was used.

I would agree that there is no need to say "is no longer available", just say "for CentOS 7 use yum"

chrisarcand: > They removed the RPM from EPEL 7 because it was old and buggy, it was never available in CentOS or RHEL core. The "right" way to install packages in CentOS 7 is yum.

<3 Thanks for clarification. Yup, @naltun let's just say use yum for CentOS 👍

chrisarcand: > Therefore, yum is the only option, no?

@naltun Yup you were right! I'm surprised it was removed but if it's really buggy and old...

PR #274, opened 2017-11-17
Advise to try npm install without sudo first

mturley:
At least on mac, I have found that avoiding `sudo npm install -g` is helpful in preventing odd permissions issues that can come up.  It is possible to keep npm entirely in userspace, so I added a note to the guide about that and removed sudo from the example commands.


mturley:
Sorry about that!  I'm used to squash-and-merge being the default on my last project.

PR #280, opened 2018-01-05
Double checking loaded records

romanblanco:
@panspagetka please take a look if the description makes sense. Thanks
  
Links: https://github.com/ManageIQ/manageiq-ui-classic/pull/3131


PanSpagetka:
Yep, I think this is good 👍

PR #282, opened 2018-01-09
Update rubocop yaml files for RuboCop 0.52.1

Fryguy:
@bdunne Please review.


.rubocop.yml
```diff
@@ -0,0 +1,4 @@
+inherit_from:
```
bdunne: Why add this here?  I don't think we need to rubocop anything in this repo, do we?

Fryguy: This is for the scripts that live in this repo.

bdunne: 👍

PR #283, opened 2018-01-10
Disable rubocop Layout/PercentLiteralDelimiters to allow %()

skateman:
Making @himdel happy :beers:

### DIFF
--- a/.rubocop_base.yml
+++ b/.rubocop_base.yml
@@ -5,6 +5,8 @@ Layout/AlignHash:
   EnforcedColonStyle: table
 Layout/ExtraSpacing:
   AutoCorrect: true
+Layout/PercentLiteralDelimiters:
+  Enabled: false
 Layout/SpaceBeforeFirstArg:
   Enabled: false
 Layout/SpaceInsideHashLiteralBraces:


himdel:
Cc @bdunne 

LGTM :+1: (I though there would be more rules though)

himdel:
Merging now, people started asking about converting it to `%w[` already :)

.rubocop_base.yml
```diff
@@ -5,6 +5,8 @@ Layout/AlignHash:
   EnforcedColonStyle: table
 Layout/ExtraSpacing:
   AutoCorrect: true
+Layout/PercentLiteralDelimiters:
```
bdunne: Isn't this a `Style` cop? https://github.com/bbatsov/rubocop/blob/v0.52.1/lib/rubocop/cop/style/percent_literal_delimiters.rb#L13-L16

Also, wouldn't we want to set a preferred delimiter rather than disabling the cop completely?
cc @Fryguy

skateman: According to http://rubocop.readthedocs.io/en/latest/cops_layout/#layoutspaceinsidepercentliteraldelimiters no :confused:

bdunne: That's a different cop.

skateman: OH :boom: sorry then :sob:

bdunne:
https://github.com/ManageIQ/guides/pull/284

PR #284, opened 2018-01-10
PercentLiteralDelimiters is Style, not Layout

bdunne:
- Re-enable the cop with () as the default


bdunne:
Fixes issues in https://github.com/ManageIQ/guides/pull/283

PR #287, opened 2018-01-18
Ignore manageiq fixture path and vendor path

imtayadeway:
Running rubocop locally on a ManageIQ plugin will check the
spec/manageiq and vendor paths unless explicitly ignored, resulting in
1000s of lines of unnecessary output.


imtayadeway:
@miq-bot assign @Fryguy

PR #289, opened 2018-02-02
use illustrating percent array literal

AlexanderZagaynov:
I don't know why the round brackets were chosen instead of the square ones, maybe there was some weighty reason?

However, I propose to use square brackets - because it looks the same as the basic array literal:
```ruby
[:a, :b] == %i[a b]
```
I believe it will make the code more readable.

### DIFF
--- a/.rubocop_base.yml
+++ b/.rubocop_base.yml
@@ -199,10 +199,11 @@ Style/NumericLiterals:
   AutoCorrect: false
 Style/PercentLiteralDelimiters:
   PreferredDelimiters:
-    '%i': ()
-    '%I': ()
-    '%w': ()
-    '%W': ()
+    default: '[]'
+    '%i': '[]'
+    '%I': '[]'
+    '%w': '[]'
+    '%W': '[]'
 Style/PerlBackrefs:
   Enabled: false
 Style/ParallelAssignment:


AlexanderZagaynov:
@bdunne @skateman @Ladas @Fryguy what do you think?

.rubocop_base.yml
```diff
@@ -199,10 +199,11 @@ Style/NumericLiterals:
   AutoCorrect: false
 Style/PercentLiteralDelimiters:
   PreferredDelimiters:
-    '%i': ()
-    '%I': ()
-    '%w': ()
-    '%W': ()
+    default: '[]'
+    '%i': '[]'
```
Ladas: I am pretty sure it was set like this in some repo

chrisarcand:
My driveby 2 cents is just that I don't have a preference between any delimiter and honestly just forget which one we use, sometimes pick `[`, and then have Rubocop yell at me. TBH I'd just allow *either* `[]` or `()` and be done with it, which also keeps us from having to update all the existing code too.  

But again, not very passionate here, do whatever. (I do use %w/i syntax a lot though)

skateman:
AFAIK we're using the `%i(a b)` form in the UI code...

Ladas:
Btw. I just use ```rubocop -a``` for this, so I am also fine having any. (I think a normal array is more readable though)

himdel:
I'm for not changing stuff unless there's a good reason.

So `()` from me :)

jrafanie:
As @chrisarcand said, I do the same thing. I generally choose whichever format I see in the same code "area" and use that.  I have no preference for either and think it's just noise for rubocop to complain about it.

himdel:
@AlexanderZagaynov can we close this? :)

PR #293, opened 2018-02-21
Add reducers example and how to dispatch action for redux

karelhala:
We were lacking definition of simple reducers created for our redux. There are two examples of simple reducers and how to dispatch new action.

### DIFF
--- a/ui/redux.md
+++ b/ui/redux.md
@@ -1,0 +1,172 @@
+### Redux
+In UI-classic we enabled [redux](https://github.com/reactjs/redux) to work with state of application so we can easilly
+ detect from where which action comes and to see the flow of data.
+
+#### Writing reducers
+To change state of application you will need to write reducer which is basically simple function which takes current 
+state and action. Result of such reducer is somehow changed state based on action.
+
+Action consists of
+ * Type of action - string which says what action has been triggered
+ * Payload - usually JS object with some data, which will be applied to state
+
+##### Simple reducer example
+We are using redux in slightly different way, so we can control which reducer is active at given time. To create new
+reducer first you have to create it and register it into redux.
+
+1. Breaking reducers into simple functions, might be a bit confusing because it doesn't use switches as any redux example
+however it's easier to test and maintain.
+
+```javascript 1.8
+/* 
+ * These two functions are required for correct registering of reducer into our redux.
+ * Please change path relative to your pack and from where you are registering reducers.
+ */
+import { addReducer, applyReducerHash } from '../miq-redux/reducer';
+
+// Constants of action types
+const FIRST_ACTION = 'first_action';
+const SECOND_ACTION = 'second_action';
+
+/* 
+ * Simple reducer function
+ * It takes state and action as parameters and changes state based on action.payload
+ * Notice the rest operator which allows changing of state with action payload.
+ */
+function firstReducerFn(state, action) {
+  return {...state, someValue: action.payload};
+}
+
+/*
+ * A bit more complex reducer, which takes two values off of action payload and applies them into state.
+ * You can notice, that we used different approach and used Object.assign for changing state.
+ */
+function secondReducerFn(state, action) {
+  const someObject = {
+    differentValue: action.payload.oneValue,
+    anotherValue: action.payload.secondValue
+  };
+  return Object.assign({}, state, someObject);
+}
+
+// This is how we say that which action type maps to which function.
+const mappedReducers = {
+  [FIRST_ACTION]: firstReducerFn,
+  [SECOND_ACTION]: secondReducerFN
+};
+
+/*
+ * This is probably the most important and also most confusing part of how to register reducer.
+ * Function addReducer helps us to add newly created reducer into our bulk of reducers which were created before.
+ * Function applyReducerHash helps us to bind state and action for each reducer which we mapped before.
+ */
+const unbind = addReducer(
+  (state, action) => applyReducerHash(mappedReducers, state, action)
+)
+
+/* Notice that we created unbind constant which is taking whatever is returned from addReducer.
+ * This little guy is just callback to our Set. Calling it will remove our reducers from all reducers, this way you can
+ * control when to remove such reducers. But don't worry it's not needed, just nice. 
+ */
+```
+
+2. Using switch as described in many redux examples. It does the same thing as example before, but it doesn't brake
+reducers into simple functions and it uses switches, which someone might like and someone not.
+```javascript 1.8
+/* 
+ * This function is required for correct registering of reducer into our redux.
+ * Please change path relative to your pack and from where you are registering reducers.
+ */
+import { addReducer } from '../miq-redux/reducer';
+
+// Constants of action types
+const FIRST_ACTION = 'first_action';
+const SECOND_ACTION = 'second_action';
+
+/*
+ * Simple redux reducer with two branches.
+ */
+function someApp(state, action) {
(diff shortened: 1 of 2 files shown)


karelhala:
@miq-bot assign @himdel 
@ohadlevy and @vojtechszocs this might interest you
@skateman you asked me how to create new reducer, well here you have two options, feel free to choose.

himdel:
@karelhala Can you also link the new article from `ui/patterns.md` (or `README.md`) please?

ui/redux.md
```diff
@@ -0,0 +1,120 @@
...
+ * Payload - usually JS object with some data, which will be applied to state
+
+##### Simple reducer example
+We are using redux in slightly different way, so we can control which reducer is active at given time. To create new
+reducer first you have to create it and register it into redux.
+
+1. Breaking reducers into simple functions, might be a bit confusing because it doesn't use switches as any redux example
+however it's easier to test and maintain.
+
+```javascript 1.8
```
skateman: Is this javascript or typescript, syntax is confusing...

ui/redux.md
```diff
@@ -0,0 +1,120 @@
...
+    anotherValue: action.payload.secondValue
+  };
+  return Object.assign({}, state, someObject);
+}
+
+// This is how we say that which action type maps to which function.
+const mappedReducers = {
+  [FIRST_ACTION] = firstReducerFn,
+  [SECOND_ACTION] = secondReducerFN
+};
```
skateman: Is this a valid syntax?

ui/redux.md
```diff
@@ -0,0 +1,120 @@
...
+// Constants of action types
+const FIRST_ACTION = 'first_action';
+const SECOND_ACTION = 'second_action';
+
+/* 
+ * Simple reducer function
+ * It takes state and action as parameters and changes state based on action.payload
+ * Notice the rest operator which allows changing of state with action payload.
+ */
+function firstReducerFn(state, action): AppState {
```
skateman: This seems like typing at the end of the function...

ui/redux.md
```diff
@@ -0,0 +1,120 @@
...
+  return {...state, someValue: action.payload};
+}
+
+/*
+ * A bit more complex reducer, which takes two values off of action payload and applies them into state.
+ * You can notice, that we used different approach and used Object.assign for changing state.
+ */
+function secondReducerFn(state, action): AppState {
```
skateman: Same here

skateman:
Also would be nice to have a convention where to put the reducers, maybe `app/javascripts/reducers` and include that path automatically to the codebase.

ui/redux.md
karelhala: It's javascript, but I forgot to remove types from couple of functions. Good point.

karelhala:
> Also would be nice to have a convention where to put the reducers, maybe app/javascripts/reducers and include that path automatically to the codebase.

Well that really does not matter, but I feel like we should discuss this so everybody will follow same pattern.

Given you are creating new pack, let's say for treeview the folder structure could be like this:
**`app/javascript/treeview/reducers/actions.js`**
```javascript
export const FIRST_TREE_ACTION = 'someAction';
export const SECOND_TREE_ACTION = 'someOtherAction';
```
**`app/javascript/treeview/reducers/reducers.js`**
```javascript
import { FIRST_TREE_ACTION, SECOND_TREE_ACTION } from './actions';

export function firstReducerFn(state, action) {
  //do something cool with state
  return state.
}

export function secondReducerFn(state, action) {
  //do something completely different with state
  return state;
}
```
**`app/javascript/treeview/reducers/index.js`**
```javascript
import { firstReducerFn, secondReducerFn } from './reducers';
import { FIRST_TREE_ACTION, SECOND_TREE_ACTION } from './actions';
import { addReducer, applyReducerHash } from '../miq-redux/reducer';

const actionsToFunctions = {
  [FIRST_TREE_ACTION]: firstReducerFn,
  [SECOND_TREE_ACTION]: secondReducerFn
}

addReducer(
  (state, action) => secondReducerFn(actionsToFunctions, state, action)
)
```

You'd then import `reducers` inside your index where you connect your application to redux.

skateman:
@karelhala well, I think that's the most important part :wink: conventions to avoid the spaghetti, so let's work on this a bit more ...

karelhala:
@Hyperkid123 FYI

Hyperkid123:
I like @karelhala approach especially replacing switch statement with object in reducer. I've been working with Redux for some time now and after several different approaches i've ended up using something like this.
```bash
├── something
├── redux
│   ├── actions
│   │   ├── index.js
│   │   ├── someActions.js
│   │   ├── otherActions.js
│   ├── actionTypes
│   │   ├── index.js
│   │   ├── someTypes.js
│   │   ├── otherTypes.js
│   ├── reducers
│   │   ├── index.js //combines reducers, adds middleware, logger etc.
│   │   ├── reducerA.js
│   │   ├── reducerB.js
│   │   ├── reducerX.js
└── something
```
I like having action types and actions in separate files because u have to use them in both actions and reducers. Having them exported in `index.js` also helped me avoid creating duplicate action types.

Actions are grouped based on the section where they are used. I haven't been using Redux in very large application like ManageIQ, so specifying if action should be inside some group or if it should be somewhere else might be a bit different. There is also a possibility to define action is some top level connected component but i don't really like that, because recently i've figured out, that connecting a lot of small components can drastically improve performance (in React, dunno about Angular). I also like to have my components as clean as possible. But maybe some separate actions file in the folder with components that should use them?

When in comes to the reducers i think they should be as specific as possible. In my opinion there should be one (or more) reducer per one (smart) component and then some "common" reducers, for instace `ui-reducer`. At first i was using one or two reducers for whole application and it was really confusing so i've ended up splitting them anyway... And since there is a bunch of forms in ManageIQ and they are all different, it doesn't really make sense to have one gigantic reducer that stores everything.

skateman:
Do we really need that `index.js` everywhere? Can't we somehow combine these things automatically? I can see the index files in the future to be 80% the same code and a point of failure if you miss to include something from it...

skateman:
Also @karelhala please join your example and the folder structure into a single thing. Having a single example and then a very similar code just separated among files is just confusing.

Hyperkid123:
@skateman `index.js` is not necessary everywhere. But with action types i think that is a good idea. You have to somehow prevent having same `actionTypes`, because it will eventyalu break your store. All reducers are listening on all dispatched actions. If u have two actions with same action type but they are designed for different reducers, reducers  catch them anyway and try to update store. Having all actions in one file or exported by one file will help you to prevent this from happening because the code won't even compile. And someone will eventually try to create action type that already exists given the high number of forms. I think its better to prevent such possibility now, the try to repair it in future. It doesn't have to be 'index.js', surely there is another and better option, but i think something like this should be there from the start.

skateman:
@Hyperkid123 my problem is that we have to explicitly include stuff just so we can use it ... and I think this is something that can be done automatically, i.e. why would you add a file into this folder structure if you don't want to include it?

Hyperkid123:
@skateman i agree

karelhala:
What's the status on this? I think having both examples would be good since we can show how to do simple reducer which is not broken into multiple files and we can show how to do complex reducer broken into multiple files.

himdel:
Looks like the only remaining problem is having to manually add stuff to `index.js` .. that's something we can probably address using some kind of `require_dir` approach, so looks like we can merge this?

Merging, we can always fix stuff later :)

PR #297, opened 2018-03-13
Update developer_setup.md

slemrmartin:
Update for Fedora 26 and PostgreSQL 9.5

### DIFF
--- a/developer_setup.md
+++ b/developer_setup.md
@@ -47,6 +47,29 @@
   sudo -u postgres psql -c "CREATE ROLE root SUPERUSER LOGIN PASSWORD 'smartvm'"
   # This command can return with a "could not change directory to" error, but you can ignore it
   ```
+#### Fedora 26 + PostgreSQL 9.5
+* Install packages
+  * Same as above
+  
+* Enable Memcached
+  * Same as above
+  
+* Configure PostgreSQL
+  * If you're installing PostgreSQL in version which is not included in distribution, directions are slightly different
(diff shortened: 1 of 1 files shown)


carbonin:
Can we make this just the default for all of Fedora here rather than adding yet another option for first-time users? I think as it stands this only makes it more difficult to get up and running.

This will now be the third documented way to install PG on Fedora. I would rather pick one then just call out the others as options.

@Fryguy @bdunne what do you think about standardizing on PGDG install of 9.5 as the "documented" way of doing things because it doesn't rely on what is shipped with a particular distro or version then just call out the other ways (SCL, distro) as options without explicit documentation?

bdunne:
Yes, I think guides should have a single recommended way to get started.  If people happen to know more about a particular component and want to go some other route, feel free, but a simple single path to getting started would be more friendly to beginners.

slemrmartin:
One more note - I installed new notebook again and I think (but not sure) that Fedora 27 has PG 9.6 preinstalled, which causes conflict and is very confusing for first-time users. 
So there is probably prerequisity: uninstall preinstalled PG server

carbonin:
I'm not sure uninstall is necessary, shouldn't you just be able to stop and disable the existing service and start the new one installed from the official PG repos?

slemrmartin:
It didn't work for me, some of next commands still wanted to use 9.6..But I'm not an expert in this area, I might miss something

djberg96:
:+1:

martinpovolny:
@slemrmartin : Closing. Please, update it to a new Fedora version and reopen if you are interested.

Sorry and thanks!

PR #298, opened 2018-03-14
Fix errors with out of date comma cops

NickLaMuro:
This prevents rubocop from running properly.

Error in question:

```
Error: The `Style/TrailingCommaInLiteral` cop no longer exists. Please use `Style/TrailingCommaInArrayLiteral` and/or `Style/TrailingCommaInHashLiteral` instead.
(obsolete configuration found in .rubocop-https---raw-githubusercontent-com-ManageIQ-guides-master--rubocop-base-yml, please update it)
```

Updated to use both cops to retain the same functionality.

Links
-----

* Changed in `rubocop` here:  https://github.com/bbatsov/rubocop/pull/5307
* Original `rubocop` rational:  https://github.com/bbatsov/rubocop/issues/3394

PR #303, opened 2018-04-23
Updates Rubocop rules to add compatibility with versions 0.53.0+

saulotoledo:
At [version`0.53.0`](https://github.com/bbatsov/rubocop/blob/3c3e315b84df45845440a25cbd71a5b99214b21d/CHANGELOG.md#0530-2018-03-05), the `TrailingCommaInLiteral` rule at Rubocop has been split into`TrailingCommaInHashLiteral` and `TrailingCommaInArrayLiteral` and no longer exists. The current version definition for Rubocop allow us to use the version `0.52.1` or superior, but it is not possible to have both versions by having that options in your `rubocop.yml` file.

Thus, we have to choose between using a version lower than `0.53.0` or using a superior version and update our rules. The current suggestion updates the rules list in order to have compatibility with `0.53.0+`.

This problem can be seen at https://github.com/ManageIQ/manageiq-ui-classic/ project, that uses the Gemfile in the core project in its build process.

Depends on:
- https://github.com/ManageIQ/manageiq/pull/17327

### DIFF
--- a/.rubocop_base.yml
+++ b/.rubocop_base.yml
@@ -233,7 +233,9 @@ Style/StringLiteralsInInterpolation:
   Enabled: false
 Style/TrailingCommaInArguments:
   Enabled: false
-Style/TrailingCommaInLiteral:
+Style/TrailingCommaInArrayLiteral:
+  Enabled: false
+Style/TrailingCommaInHashLiteral:
   Enabled: false
 Style/TrailingUnderscoreVariable:
   Enabled: false


saulotoledo:
@miq-bot assign @bdunne

saulotoledo:
@miq-bot add_reviewer @bdunne

djberg96:
:+1:

himdel:
@saulotoledo The problem in ui-classic you're referencing is likely a Scrutinizer issue.

Scrutinizer is supposed to respect the rubocop version in the gemfile, but currently doesn't. (@skateman opened a bug but it doesn't seem publicly visible.)

djberg96:
@himdel is there a way to support both?

himdel:
@djberg96 Not if we want miq-bot to work :)

saulotoledo:
@himdel @djberg96 Yes, this is not supported in `0.52.1`, so I updated the version at https://github.com/ManageIQ/manageiq/pull/17327 
I also agree with @himdel that we cannot support both, unless we remove those rules, and don't think that this is a good idea.

@bdunne, the linked PR is https://github.com/ManageIQ/manageiq/pull/17327 . It should be approved with this one. I added the version `0.53.0` because is the smaller version number that supports these rules, but I can upgrade that if you want.

@himdel Yes, I started updating these rules and the version because of the problem at Scrutinizee. After some investigation I agree that this is a problem at Scrutinizer, as you said (@skateman, do you have a link for that bug?), but I think we can solve in our side by just upgrading the version and the rules. https://github.com/ManageIQ/manageiq/pull/17327 contais the version upgrade.

bdunne:
@saulotoledo Every repo that uses rubocop in the ManageIQ org needs to be updated at the same time.  And the changes need to be backported too.

saulotoledo:
@bdunne Sorry, I missed the other repositories. Below is a detailed list with all projects using these configurations here, but that have their own Gemfiles. I can create the PRs in the other repositories if you want. Just let me know if you and @Fryguy agree and I will start creating the PRs.

**Updated PRs:**

- https://github.com/ManageIQ/manageiq/pull/17327
- https://github.com/ManageIQ/miq_bot/pull/423

**The following projects should be updated (I will create the PRs if you agree):**

- https://github.com/ManageIQ/pg-pglogical
- https://github.com/ManageIQ/vmware_web_service

**The following projects do not have the Rubocop version at Gemfile/gemspec:**

- https://github.com/ManageIQ/linux_admin
- https://github.com/ManageIQ/manageiq-appliance_console
- https://github.com/ManageIQ/manageiq-postgres_ha_admin
- https://github.com/ManageIQ/manageiq-schema
- https://github.com/ManageIQ/manageiq-smartstate

**The following projets to not have Rubocop in the Gemfile/gemspec:**

- https://github.com/ManageIQ/activerecord-id_regions
- https://github.com/ManageIQ/amazon_ssa_support
- https://github.com/ManageIQ/awesome_spawn
- https://github.com/ManageIQ/azure-armrest
- https://github.com/ManageIQ/log_decorator
- https://github.com/ManageIQ/manageiq-appliance
- https://github.com/ManageIQ/manageiq-gems-pending
- https://github.com/ManageIQ/manageiq-messaging
- https://github.com/ManageIQ/manageiq-network_discovery
- https://github.com/ManageIQ/more_core_extensions
- https://github.com/ManageIQ/ovirt

**The following projects do not have a Gemfile:**
https://github.com/ManageIQ/manageiq-appliance-build

Also, this PR is a duplicate of https://github.com/ManageIQ/guides/pull/298 . Should we move the discussion to there? Or keep them here?

saulotoledo:
@bdunne @himdel @djberg96 Should we continue working here?

skateman:
Well, this might solve our problems with CodeClimate and Scrutinizer...so would be nice to get these in. cc @martinpovolny

saulotoledo:
@skateman, @martinpovolny Just let me know and I start working on the listed repositories.

martinpovolny:
I don't have the rights to do anythig with the PRs above. The one who can do the most here is @Fryguy .

saulotoledo:
@martinpovolny @bdunne This PR is no longer necessary due to https://github.com/ManageIQ/guides/pull/351/.

I had similar fixes on several other PRs in this list that were automatically closed by @miq-bot due to waiting for more than 6 months. I can update and reopen them all. Let me know if you are still interested.

Fryguy:
Thanks for the updates @saulotoledo !

saulotoledo:
@Fryguy , should I reopen the other PRs (and update the version to match the same on #351 )?

Fryguy:
@saulotoledo So, here's the general way I think it should work, IMO.  The goal is to minimize having to update dozens of repos whenever we make a change on the bot

- ManageIQ/manageiq **should** reference rubocop in the Gemfile with the exact version (and also rubocop-performance now)
- Any plugin repo (e.g. manageiq-ui that uses ManageIQ/manageiq as a spec/manageiq reference **should not** reference rubocop in their gemspec nor Gemfile.  This is because it will be pulled in automatically via the spec/manageiq reference.
- Any non-plugin repo (like standalone gems) **should** have a development reference in their gemspec to rubocop and rubocop-performance, **but should not** specify a version.  Otherwise it's going to be a pain every time we have to bump core.  For most of these gems, being a little off should be ok.
- All repos should have a copy of the same .rubocop.* files and changes to the .codeclimate.yml file - here are the templates:
  - https://github.com/ManageIQ/manageiq/blob/master/lib/generators/manageiq/plugin/templates/.rubocop.yml
  - https://github.com/ManageIQ/manageiq/blob/master/lib/generators/manageiq/plugin/templates/.rubocop_cc.yml
  - https://github.com/ManageIQ/manageiq/blob/master/lib/generators/manageiq/plugin/templates/.rubocop_local.yml  (this file may have local changes, and that's ok...if the file doesn't exist though, it should be created and blank).
  - https://github.com/ManageIQ/manageiq/blob/master/lib/generators/manageiq/plugin/templates/.gitignore#L1
  - https://github.com/ManageIQ/manageiq/blob/master/lib/generators/manageiq/plugin/templates/.codeclimate.yml#L3-L8 and https://github.com/ManageIQ/manageiq/blob/master/lib/generators/manageiq/plugin/templates/.codeclimate.yml#L27-L30
- https://github.com/ManageIQ/manageiq-appliance-build is just weird and we should open an issue there to fix the Gemfile location as well as other oddities.  For now, just leave it as is.

I think that should cover all of the repos you mentioned.  So, *some* of your PRs may need to be reopened or tweaked, but if we can align with this, I think in the future it will be a lot easier to update the bot more frequently.

PR #304, opened 2018-04-25
Add `gem` to the exclusion list

bdunne:
Based on https://github.com/ManageIQ/manageiq-appliance/pull/188#issuecomment-384318895

cc @carbonin


jrafanie:
👏 🙇‍♀️ :shipit:

PR #305, opened 2018-04-26
Add information about report data API

karelhala:
### New section
New section about report data API is added, which describes each action and examples of how to use these actions.

### DIFF
--- a/ui/report_data_api.md
+++ b/ui/report_data_api.md
@@ -1,0 +1,221 @@
+### Report data API
+Report data is component which is reponsible for showing data inside Grid, Tile and List (refered to as GTL). This compont shows a lot of data so it comes with API along side of it. This API is for working with GTL and both changing it's state and getting data out of it.
+
+#### Sending data
+To initiate any action you just simply call specific function called `sendDataWithRx` which is hooked to `Rx.js`'s Subject and forwards these messeges to corresponding controller.
+
+You have to specify 3 attributes when comunicating with report data API
+1) controller - `reportDataController`, constant and should not be changes
+2) action - what action should be initiated
+3) data - if action requires some data pass them via this attribute
+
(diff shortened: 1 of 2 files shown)


karelhala:
@miq-bot assign @himdel 
@izapolsk ping this documentation might be useful for you so you don't have to remember each action and what to send there.

ui/report_data_api.md
```diff
@@ -0,0 +1,221 @@
+### Report data API
+Report data is component which is reponsible for showing data inside Grid, Tile and List (refered to as GTL). This compont shows a lot of data so it comes with API along side of it. This API is for working with GTL and both changing it's state and getting data out of it.
```
himdel: s/compont/component/

ui/report_data_api.md
```diff
@@ -0,0 +1,221 @@
...
+
+#### Sending data
+To initiate any action you just simply call specific function called `sendDataWithRx` which is hooked to `Rx.js`'s Subject and forwards these messeges to corresponding controller.
+
+You have to specify 3 attributes when comunicating with report data API
+1) controller - `reportDataController`, constant and should not be changes
```
himdel: change**d**?

ui/report_data_api.md
```diff
@@ -0,0 +1,221 @@
...
+})
+```
+Notice that data is always array even if you are sending single value always pass it as array.
+
+Some actions will store data in `ManageIQ.qe.gtl.result`, it's either direct result (string, number) or object which contains item or items, based on action.
+
+#### Actions
+
+##### `select_item`
+Action which selects item based on it's id. Data should be array with first attribut ID of item and second `true|false` to select|deselect item.
```
himdel: its

himdel:
LGTM, just a couple of typos

PR #307, opened 2018-05-14
Add information about shared component API

karelhala:
Shared component API guide to show how to create and use and simple list of interaction with such component.

### DIFF
--- a/ui/register_react_component.md
+++ b/ui/register_react_component.md
@@ -1,0 +1,125 @@
+### Register react component
+In ManageIQ project there is a way of sharing components, getting their state and control their flow by something which is called ManageIQ component.
+
+This registry and usage consists of three steps, define blueprint that holds the react component with its props. Then you will define it inside shared components (use blueprint created in step 1). And when someone decides to use such component he will call function newInstance with custom props and element to which such component is bound to.
+
+This way using such component is abstract for user, he doesn't know if such component was written in pure JS, React, Angular or whatever.
+
+#### 1. Define react component
+This is your first step of creating shared abstract component. You may use our custom function which fits react, should you choose angular, Vue.js or whatever you'll have to write your own funcion.
+
+When using react blueprint function, just wrap your component (or HTML) with `addReact` function and as argument use callback which will be called with props. These props will be passed when new isntance of this component is created.
+
+```JSX
+import React from 'react';
+import addReact from 'helpers';
+import { YourComponent } from 'your-components/custom-component';
+
+addReact('YourComponentName', YourComponent);
+```
+
+#### 2. New instance
+When you want to create new instance of recently created component just call `newInstance` function with these arguments
+* name - name of the component you wish to use.
+* element - DOM element where this new instance of component should be rendered.
+* props - object with props that will be mapped to your components (used when creating blueprint)
+
+Props are optional, if you do not pass props object the component factory will look for data attribute of DOM element.
+```JS
+import { componentFactory } from 'helpers';
+
+componentFactory('YourComponentName', '#selector');
+```
+
+This will mount new component into DOM element which might look like this.
+```haml
+#selector{"data-something" => "some string",
+          "data-another"   => "another string"}
+```
+
+To trigger the component creation either put the JS code into `-common` package if this is aplicable to all pages. Or create new pack which might look like this.
+```JS
+import { componentFactory } from 'helpers';
+
+document.addEventListener("DOMContentLoaded", function(event) { 
+  componentFactory('YourComponentName', '#selector');
+});
+```
+
+And to load such new pack use `= javascript_pack_tag 'manageiq-ui-classic/your-pack'` inside HAML with `selector` DOM element.
+
+### Shared component API
+Abstract shared components has API controlling active components, this way you can define and create new component, later you can access to all active components on screen and use its shared interface to interact with it.
+
+#### Define
+When you want to define new component to be used by someone call define function with name of such component and its blueprint.
+```JS
+import define from 'registry';
+
+define('YourComponentName', componentBlueprint);
+```
+
+#### Get definition
+To fetch definition of component you created you can call specific function. You will get name of such component and its blueprint. With such blueprint you can call functions to update, create or remove it.
+```JS
+import getDefinition from 'registry';
+
+getDefinition('YourComponentName');
+```
+
+#### New instance
+To create new instance and mount it to DOM element call function `newInstance`. This function not only creates new instance it also gives you object over which you can update its pros, update it or destroy it from component's drawer. Inside this object you will also find comeponent's instance ID and should the component choose its interact property for side channel communication.
+
+When creating new instance you have option to pass multiple parameters:
+* name - component's name under which component was defined
+* props - default props for the component
+* element - DOM element under which this component should be rendered (optional)
+```JS
+import newInstance from 'registry';
+
+newInstance('YourComponentName', {}, element);
+```
+
+#### export instance
+To receive component's instance from registery you can call function `getInstance`. You will get full information about this instance, ways to interact with it and s name and ID.
+```JS
+import getInstance from 'registry';
+
+getInstance('YourComponentName', componentID);
+```
+
+#### Is defined
+Check if component with some name is defined in drawer. Great for checking if component's name is available.
+```JS
+import isDefined from 'registry';
+
+isDefined('YourComponentName');
+```
+
+### Component instance API
+When you create new instance or search for instance of component you will receive object over which you can interact with this instance and observe its properties.
+
+#### ID
+Instance's ID under which its created.
+
+#### interact
+Should the component chooses it can supply object over which its possible side channel communication.
+
+#### props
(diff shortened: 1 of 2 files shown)


ui/register_react_component.md
```diff
@@ -0,0 +1,98 @@
...
+This is your first step of creating shared abstract component. You may use our custom function which fits react, should you choose angular, Vue.js or whatever you'll have to write your own funcion.
+
+When using react blueprint function, just wrap your component (or HTML) with `addReact` function and as argument use callback which will be called with props. These props will be passed when new isntance of this component is created.
+
+```JSX
+import React from 'react';
+import addReact from 'helpers';
+import { YourComponent } from 'your-components/custom-component';
+
+addReact('YourComponentName', props => <YourComponent {...props} />);
```
martinpovolny: why don't just call
```
addReact('YourComponentName', YourComponent);
```
?

ui/register_react_component.md
```diff
@@ -0,0 +1,98 @@
+### Register react component
+In ManageIq project there is a way of sharing components, getting their state and control their flow by something which is called ManageIq component.
```
himdel: ManageIQ, the q is big ;)

ui/register_react_component.md
```diff
@@ -0,0 +1,98 @@
...
+```
+
+### Component instance API
+When you create new instnace or search for instance of component you will receive object over which you can interact with this instance and observe it's properties.
+
+#### ID
+Instance's ID under which it's created.
+
+#### interact
+Should the component chooses it can supply object over which it's possible side channel communication.
```
martinpovolny: s/chooses/choose/

which/that?

"possible side channel communication" -- verb missing?

ui/register_react_component.md
```diff
@@ -0,0 +1,98 @@
...
+
+#### props
+Instance's props object containing public properties.
+
+#### update
+This function helps you to update component's public props. You can add new prop or update existing prop.
+
+#### destroy
+To destroy instance and remove it from DOM (no point of return) simply call this function.
```
martinpovolny: no way back?

ui/register_react_component.md
```diff
@@ -0,0 +1,98 @@
...
+#### Is defined
+Check if component with some name is defined in drawer. Great for checking if component's name is available.
+```JS
+import isDefined from 'registry';
+
+isDefined('YourComponentName');
+```
+
+### Component instance API
+When you create new instnace or search for instance of component you will receive object over which you can interact with this instance and observe it's properties.
```
martinpovolny: s/instnace/instance

ui/register_react_component.md
```diff
@@ -0,0 +1,98 @@
+### Register react component
+In ManageIq project there is a way of sharing components, getting their state and control their flow by something which is called ManageIq component.
+
+This registry and usage consists of three steps, define blueprint which holds the react component with it's props. Then you will define it inside shared components (use blueprint created in step 1). And when someone decides to use such component he will call function newInstance with custom props and element to which such component is bound to.
```
himdel: it's = it is
its = possessive 

... in many places :)

ui/register_react_component.md
```diff
@@ -0,0 +1,98 @@
...
+
+getDefinition('YourComponentName');
+```
+
+#### New instance
+To create new instance and mount it to DOM element call function `newInstance`. This function not only creates new instance it also gives you object over which you can update it's pros, update it or destroy it from component's drawer. Inside this object you will also find comeponent's instance ID and should the component chooses it's interact property for side channel communication.
+
+When creating new instance you have option to pass multiple parameters:
+* name - component's name under which component was defined
+* props - default props to component
```
martinpovolny: default props for the component

ui/register_react_component.md
```diff
@@ -0,0 +1,98 @@
...
+#### 2. New instance
+When you want to create new instance of recently created component just call newInstance function with these arguments
+* name - name of component you wish to use.
+* props - object with props which will be mapped to your components (used when creating blueprint)
+* element - DOM element where this new instance of component should be rendered.
+```JS
+import newInstance from 'registry';
+
+const props = {};
+const element = document.getElementById('element-id');
```
himdel: no getElementById, `newInstance` should take a normal selector

(or should this be using `componentMarkup` instead?)

ui/register_react_component.md
martinpovolny: s/which holds/that holds

ui/register_react_component.md
```diff
@@ -0,0 +1,98 @@
...
+```JSX
+import React from 'react';
+import addReact from 'helpers';
+import { YourComponent } from 'your-components/custom-component';
+
+addReact('YourComponentName', props => <YourComponent {...props} />);
+```
+
+#### 2. New instance
+When you want to create new instance of recently created component just call newInstance function with these arguments
```
martinpovolny: `newInstance` (backticks missing)

ui/register_react_component.md
```diff
@@ -0,0 +1,98 @@
...
+* name - name of component you wish to use.
```
martinpovolny: name of the component

ui/register_react_component.md
```diff
@@ -0,0 +1,98 @@
...
+* props - object with props which will be mapped to your components (used when creating blueprint)
```
martinpovolny: s/props which will/props that will/

himdel:
Also this is missing any mention of `ManageIQ.component.*` - we probably should import & use in es6 code, but any es5 code needs to know about the functions being available on `ManageIQ.component`.

himdel:
This should probably also include a way to get redux running...

Right now, using newInstance/componentMarkup, it's impossible to use redux, unless the original component is defined with <Provider...> around it.

Hyperkid123:
@karelhala can you add [the component factory usage](https://github.com/ManageIQ/manageiq-ui-classic/pull/3931)?

ui/register_react_component.md
karelhala: New instance expects mount to element, because we don't know how user gets such element either trough getElement, trough react prop or whatever. @Hyperkid123 is writing helper method which will be shortcut to these commands.

karelhala:
@himdel @martinpovolny thanks for the review :+1: . Fixed all of your comments and added section about import mapping so when we use ES5 we know where to look for.

ui/register_react_component.md
himdel: What I'm saying is - the default method to instantiate a component must follow the same rules as all the other methods (like `miq_bootstrap`).

So.. if this documents `newInstance` as something you would want to use, we need to change it to take selectors.

If this documents `componentMarkup` as the thing to use, and mentions `newInstance` only for the cases where you're doing stuff manually, we don't need to change it.

But.. right now, it appears like `newInstance` is the preferred way, possibly not what you intended?

ui/register_react_component.md
```diff
@@ -0,0 +1,107 @@
...
+This is your first step of creating shared abstract component. You may use our custom function which fits react, should you choose angular, Vue.js or whatever you'll have to write your own funcion.
+
+When using react blueprint function, just wrap your component (or HTML) with `addReact` function and as argument use callback which will be called with props. These props will be passed when new isntance of this component is created.
+
+```JSX
+import React from 'react';
+import addReact from 'helpers';
+import { YourComponent } from 'your-components/custom-component';
+
+addReact('YourComponentName', YourComponent);
```
himdel: Looks like we may need to add something like...

If your component is not a pure function but a react class, you will need to call `addReact('YourComponentName', (props) => <YourComponent {...props}>);` instead.

(But, maybe it would be better to fix that, let's talk tomorrow about how :))

martinpovolny: Yes, I hit that today.

ui/register_react_component.md
```diff
@@ -0,0 +1,107 @@
...
+```
+
+#### 2. New instance
+When you want to create new instance of recently created component just call `newInstance` function with these arguments
+* name - name of the component you wish to use.
+* props - object with props that will be mapped to your components (used when creating blueprint)
+* element - DOM element where this new instance of component should be rendered.
+```JS
+import componentFactory from 'helpers';
```
Hyperkid123: `componentFactory` is not a default export, should be `import { componentFactory } from ...`

ui/register_react_component.md
```diff
@@ -0,0 +1,107 @@
...
+
+const props = {};
+componentFactory('YourComponentName', props, 'selector');
```
Hyperkid123: Order of parameters is wrong, should be `componentFactory('YourComponentName', 'selector', props)`, also props is not mandatory and you can pass then via. data atribute of them DOM element

himdel:
@karelhala The "New Instance" section doesn't actually show anything about using the component in manageiq.

This needs an example of a rails view using a react component with params :).

ui/register_react_component.md
himdel: => https://github.com/ManageIQ/manageiq-ui-classic/pull/3941

PR #310, opened 2018-05-29
Set Style/RescueStandardError to implicit

Fryguy:
Most of the code base is already in implicit mode, but rubocop changed
their default to explicit.  This moves it back for consistency as well
as my personal preference :)

https://www.rubydoc.info/gems/rubocop/0.52.1/RuboCop/Cop/Style/RescueStandardError

Please vote on what you prefer with 👍  or 👎  on this comment  (:+1: means you want implicit like this PR; :-1: means you want explicit as per the rubocop default)

### DIFF
--- a/.rubocop_base.yml
+++ b/.rubocop_base.yml
@@ -224,6 +224,8 @@ Style/RegexpLiteral:
   Enabled: false
 Style/RescueModifier:
   AutoCorrect: false
+Style/RescueStandardError:
+  EnforcedStyle: implicit
 Style/SignalException:
   EnforcedStyle: only_raise
 Style/SingleLineBlockParams:


bdunne:
I think I prefer `explicit` to encourage people to rescue only the exceptions they are expecting.

Fryguy:
Strangely, the Ruby Style Guide itself, says that `rescue => err` is good: https://github.com/bbatsov/ruby-style-guide#no-blind-rescues

There doesn't seem to be an entry in the style guide for this particular cop.

Fryguy:
Opened a PR on ruby-style-guide about the inconsistency - https://github.com/bbatsov/ruby-style-guide/issues/719

djberg96:
I'm used to implicit, but don't have any objections to making it explicit. Just means lots of cleanup.

PR #314, opened 2018-06-07
Start of the UI plugin guide

himdel:
Cc @martinpovolny 

Thanks to the whole v2v effort, we now support UI plugins. But looks like they are not documented at all yet :).

We should fix this, but for now, I mostly needed a place to write down the list of libraries the plugin can always expect to be available. Hence this PR.

(We'll probably need to add a few more, like react & react-dom.)


priley86:
it would be nice if we could add some of the babel plugins/transforms to this list. I agree that these already noted are good candidates though.

PR #316, opened 2018-06-19
InventoryCollection definition 

slemrmartin:
**Issue** https://github.com/ManageIQ/manageiq/issues/17396

Description of ```ManagerRefresh::Inventory::Persister#add_collection()``` and builder classes for definition of InventoryCollection

Cc @agrare

### DIFF
--- a/providers/dev-guide.md
+++ b/providers/dev-guide.md
@@ -257,10 +257,6 @@ anything wrong with the migration, one can go back and undo it by `rake db:rollb
 
 [documentation](http://edgeguides.rubyonrails.org/active_record_migrations.html)
 
-## Defining Inventory
-
-Provider's inventories are defined by classes named **Persister**
-
 ## Handling the Refresh Logic and Saving to the DB
 
 Refresh logic mostly happens in a `refresh_parser.rb` class. Parsed entities are then processed by core `ems_refresh.rb`.
--- a/providers/dev-guide.md
+++ b/providers/dev-guide.md
@@ -257,12 +257,25 @@ anything wrong with the migration, one can go back and undo it by `rake db:rollb
 
 [documentation](http://edgeguides.rubyonrails.org/active_record_migrations.html)
 
+## Defining Inventory
+
+Provider's inventories are defined by classes named **Persister**
+
 ## Handling the Refresh Logic and Saving to the DB
 
 Refresh logic mostly happens in a `refresh_parser.rb` class. Parsed entities are then processed by core `ems_refresh.rb`.
 The logic in this class has
 quite strong assumptions on the data being stored. It assumes that it has the tree structure and each entity contains its kids as a nested hash.
 If you are able to achieve that structure in the `refresh_parser.rb`, you are halfway done. Otherwise, good luck :]
+
+### Defining Inventory
+
+Basically, refresh process consists of Collector, Persister and Parser classes (in most providers).
+When Collector loads data from external provider (like AWS), Persister defines structure of Inventory by InventoryCollection objects.  
+Inventory is then saved to VMDB (ManageIQ (also CFME) app database). Each inventory belons to certain ActiveRecord model.
+Parser maps collected data to these inventories.
+
+Persister's InventoryCollection definition is described [there](persister/inventory_collections.md).
 
 ## Registering the Features in for RBAC
 If everything is as it should be, but you still can't see anything in the UI, this may be the purpose. MiQ has the RBAC model
--- a/providers/dev-guide.md
+++ b/providers/dev-guide.md
@@ -268,7 +268,7 @@ If you are able to achieve that structure in the `refresh_parser.rb`, you are ha
 
 Providers which use graph refresh consists of Collector, Persister and Parser classes.
 When the Collector loads data from external provider, the Persister defines the inventory structure based on InventoryCollection objects. 
-Inventory is then saved to VMDB (ManageIQ (also CFME) app database). Each inventory collection is mapped to an ActiveRecord model.
+Inventory is then saved to VMDB (app database). Each inventory collection is mapped to an ActiveRecord model.
 Parser maps data collected from the provider to the common format defined by the ActiveRecord model.
 
 Persister's InventoryCollection definition is described [there](persister/inventory_collections.md).
--- a/providers/dev-guide.md
+++ b/providers/dev-guide.md
@@ -266,10 +266,10 @@ If you are able to achieve that structure in the `refresh_parser.rb`, you are ha
 
 ### Defining Inventory
 
-Basically, refresh process consists of Collector, Persister and Parser classes (in most providers).
-When Collector loads data from external provider (like AWS), Persister defines structure of Inventory by InventoryCollection objects.  
-Inventory is then saved to VMDB (ManageIQ (also CFME) app database). Each inventory belons to certain ActiveRecord model.
-Parser maps collected data to these inventories.
+Providers which use graph refresh consists of Collector, Persister and Parser classes.
+When the Collector loads data from external provider, the Persister defines the inventory structure based on InventoryCollection objects. 
+Inventory is then saved to VMDB (ManageIQ (also CFME) app database). Each inventory collection is mapped to an ActiveRecord model.
+Parser maps data collected from the provider to the common format defined by the ActiveRecord model.
 
 Persister's InventoryCollection definition is described [there](persister/inventory_collections.md).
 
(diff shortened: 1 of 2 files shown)


providers/dev-guide.md
```diff
@@ -264,6 +264,15 @@ The logic in this class has
 quite strong assumptions on the data being stored. It assumes that it has the tree structure and each entity contains its kids as a nested hash.
 If you are able to achieve that structure in the `refresh_parser.rb`, you are halfway done. Otherwise, good luck :]
 
+### Defining Inventory
+
+Basically, refresh process consists of Collector, Persister and Parser classes (in most providers).
```
agrare: Don't need `Basically`, and I would suggest instead of `(in most providers)` how about `Providers which use graph refresh consist of Collector ...`

providers/dev-guide.md
```diff
@@ -264,6 +264,15 @@ The logic in this class has
...
+When Collector loads data from external provider (like AWS), Persister defines structure of Inventory by InventoryCollection objects.
```
agrare: Should be `When the Collector`, not sure `(like AWS)` adds anything I think any developer reading this knows what a provider is, how about `the Persister defines the inventory structure based on InventoryCollection objects`

providers/dev-guide.md
```diff
@@ -264,6 +264,15 @@ The logic in this class has
 quite strong assumptions on the data being stored. It assumes that it has the tree structure and each entity contains its kids as a nested hash.
 If you are able to achieve that structure in the `refresh_parser.rb`, you are halfway done. Otherwise, good luck :]
 
+### Defining Inventory
+
+Basically, refresh process consists of Collector, Persister and Parser classes (in most providers).
+When Collector loads data from external provider (like AWS), Persister defines structure of Inventory by InventoryCollection objects.  
+Inventory is then saved to VMDB (ManageIQ (also CFME) app database). Each inventory belons to certain ActiveRecord model.
```
agrare: Maybe just `Inventory is then saved to the database`?

agrare: > Each inventory belons

typo should be belongs

agrare: How about `Every inventory collection is mapped to an ActiveRecord model`?

providers/dev-guide.md
```diff
@@ -264,6 +264,15 @@ The logic in this class has
...
+Parser maps collected data to these inventories.
```
agrare: How about `Parser maps data collected from the provider to the common format defined by the ActiveRecord model` or something like that?

providers/persister/inventory_collections.md
```diff
@@ -0,0 +1,292 @@
+# Inventory Collection
+2018/06/19
+
+[back](../dev-guide.md)
+
+**Inventory Collection** (```ManagerRefresh::InventoryCollection``` - next called **"IC"**) is an object used by Refresh process.
```
agrare: > next called **"IC"**

Subsequently called **"IC"** sounds a bit better, next sounds like just one time after

providers/persister/inventory_collections.md
```diff
@@ -0,0 +1,292 @@
+# Inventory Collection
+2018/06/19
+
+[back](../dev-guide.md)
+
+**Inventory Collection** (```ManagerRefresh::InventoryCollection``` - next called **"IC"**) is an object used by Refresh process. 
+It represents properties of inventory like **VM**, **Hardware**, **Operating system**, **Host** etc.
```
agrare: > It represents properties of inventory like

How about `It represents types of inventory such as ...`?  When I think of properties of inventory I think of e.g. cpus/memory not whole classes of objects

providers/persister/inventory_collections.md
```diff
@@ -0,0 +1,292 @@
...
+
+Each IC is defined by persister (```ManagerRefresh::Inventory::Persister```), stored in it (in hash ```@collections```) and accessible as persister's method. Its name equals IC's association property (will be described later). One persister usually contains many IC definitions.
```
agrare: > stored in it (in hash

Think `they are stored in a hash called @collections` is better

agrare: > Its name equals IC's

should be `It's` and `is equal to the IC's`

providers/persister/inventory_collections.md
```diff
@@ -0,0 +1,292 @@
...
+
+## Defining IC
```
agrare: How about `Defining an Inventory Collection`

providers/persister/inventory_collections.md
```diff
@@ -0,0 +1,292 @@
...
+We can split this process into following areas:
+
+- Persister's helper
+  - includes *add_collection()* interface common to all persisters derived from ```ManagerRefresh::Inventory::Persister```
+- InventoryCollection Builder
+  - set of classes used for definition of IC properties (through add_collection())
+- Common definitions
+  - Definitions of ICs shared accross providers divided by provider's type (cloud/infra etc.)
+- Provider specific definitions
+  - Recommendations how to structure definitions specifical to provider and type of refresh
```
agrare: :+1: nice summary

providers/persister/inventory_collections.md
```diff
@@ -0,0 +1,292 @@
...
+
+Entry point for creating IC definition. It defines parameters:
+```ruby
+add_collection(builder_class, collection_name, extra_properties = {}, settings = {}, &block)
+```
+which we'll see more detailed in this chapter.
+
+It creates ```ManagerRefresh::InventoryCollection```  and assigns it to **persister's @collections[:vms]**
+
+Method can throw exception **`ManagerRefresh::InventoryCollection::Builder::MissingModelClassError`**
```
agrare: Maybe throw in a quick explanation about why you might get a missing model class error and how to fix it?

providers/persister/inventory_collections.md
```diff
@@ -0,0 +1,292 @@
...
+
+#### - builder_class [*mandatory*]
+
+There are 5 builder classes which groups ICs of the same manager(**TODO**, right?). Because they're used as parameter of ```add_collection()``` they are wrapped to methods with short name (persister's methods):
```
agrare: Maybe be better to say "there is a builder class for every manager type, e.g. `cloud, infra, network, ...`

providers/dev-guide.md
slemrmartin: ok, I wrote `like AWS` because it seems there is unclear terminology, what provider is. 
One time provider is EMS in our database, another time AWS/Azure/etc. is meant.

providers/dev-guide.md
slemrmartin: I wrote VMDB intentionally because all documentations uses `VMDB` when talking about our database. 
So I rather write both / `VMDB` and `app database`) because at the beginning I was confused what vmdb is. 
I think there is no other place in this doc where VMDB term is explained

slemrmartin: Also relation between ManageIQ and CFME is explained nowhere (this is not the best place but at least something for new developer)

slemrmartin: others :+1:

providers/persister/inventory_collections.md
slemrmartin: I mean IC defines properties of inventory. As an inventory I mean VM, HW etc.
--> my lng skills :)

slemrmartin:
Cc @martinpovolny please merge after @agrare approves

providers/dev-guide.md
agrare: > Also relation between ManageIQ and CFME is explained nowhere 

Actually this is intentional, these are upstream docs/guides and should make no mention of CFME

providers/persister/inventory_collections.md
agrare: Haha no worries :)

martinpovolny:
@slemrmartin : waiting for you to resolve latest @agrare 's comments.

:+1: And THANK YOU for writing the docs :+1: !

providers/persister/inventory_collections.md
```diff
@@ -0,0 +1,292 @@
...
+    :vendor   => "amazon",
+    :name     => "unknown",
+    :location => "unknown"
+  } # defined in ManagerRefresh::InventoryCollection::Builder::Shared#vms
+}
+```
+
+## Conclusion
+
+Definition of InventoryCollection objects in persisters ensures ```add_collection()``` method.
```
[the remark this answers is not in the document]
slemrmartin: Yes, I mean what you wrote, of course :))

PR #318, opened 2018-07-12
[WIP] Add manageiq-api developer guides

NickLaMuro:
As the docs say added in this pull request say, this is a guide to describe the current architecture and code structure of the current api, found in the ManageIQ/manageiq-api repo.

This is intended to be a living document that can be added to as things change, but should be geared toward getting developers familiar with the codebase and concepts of the current API.  It will not fully describe specifics of certain resources, just the general concepts that can be applied universally to specific resources.


juliancheal:
One suggestion, would be to turn some of the code paths into actual links.

NickLaMuro:
@juliancheal Yeah, I was thinking about that myself, but also adding code snippets in as well in case things change... though I guess if I just make sure to use github permalinks (point to a specific SHA), that shouldn't be too much of an issue.

Probably will be a little from column A and a little from column B.  Good suggestion!

juliancheal:
@NickLaMuro 

Yes for example:

> The first part of manageiq-api codebase that you should familiarize yourself with is the [app/controllers/api/base_controller.rb](https://github.com/ManageIQ/manageiq-api/blob/master/app/controllers/api/base_controller.rb)

But like you say with a more perma-link.

PS. While I'm here I noticed you spelt controller wrong `base_constroller.rb` I did a spelling!

abellotti:
why not put this in the manageiq-api repo and reference it from here.
kind'a a peer to https://github.com/ManageIQ/manageiq-api/blob/master/CONTRIBUTING.md

NickLaMuro:
@abellotti 

> why not put this in the manageiq-api repo ...
> kind'a a peer to https://github.com/ManageIQ/manageiq-api/blob/master/CONTRIBUTING.md

A.  Because I missed that was there (like a derp...), and looks very similar to blog post (that I also missed until recently... like a derp).
B.  I just wanted to find somewhere that didn't seem like it was going to mess with folks trying to code review legit additions, where this was just a brain dump of what I was learning as I read through the code base.  Finding the proper home was a step for the future.

Neither are "good" reasons, just reasons.  I am happy to add these additions there if you want and if you think they would be helpful to a larger audience.

Specifically to:

> ... and reference it from here.

I think there is some value in having the `guides` repo point users in the proper direction to the `manageiq-api` repo, similar to how we have setup guides for ManageIQ/manageiq, ManageIQ/manageiq-ui-classic, and the providers in the repo as well, all in one place.  That said, I am also for keeping this in the `manageiq-api` repo, since that would be the group maintaining it as well and PRs adding new general functionality can also be required to update the `CONTRIBUTING.md` (or maybe `ARCHITECTURE.md` doc) before merging.  Maybe we can just bring this in somehow to this (the `ManageIQ/guides`) repo, or have the `ManageIQ/manageiq.org` repo pull it in when building the site.

Just some thoughts, but happy to put this where you feel it makes sense.

abellotti:
❤️ for ARCHITECTURE.md

api/manageiq-api.md
```diff
@@ -0,0 +1,217 @@
...
+  end
+end
+```
+
+This is because it uses, without modifications, the code from the
+`Api::BaseController`.  Now this is basic inheritance, so there is nothing
+interesting here of note, accept for the fact that not every method defined in
+the `Api::BaseController` is available as a route in the
+`Api::ServersController`:
+
```
jvlcek: @NickLaMuro I think before showing the user of `bin/rake routes` it would be valuable to mention that this command must be run in the manageiq repo. See: https://github.com/ManageIQ/manageiq-api#development

Perhaps:
Development of the manageiq-api code base is done by pointing your local clone of the manageiq repo
and pointing it to your local branch manageiq-api repo. How to do this is described in [this section ](https://github.com/ManageIQ/guides/blob/master/developer_setup/plugins.md#dependency-on-a-local-gem) of the general developer setup guide.

Once this is set up it will be possible to query the established routes from by running the following commands from the top level directory of your manageiq repository.

NickLaMuro:
Note to those following along at home:  I was talking with @jvlcek out of band and referenced this PR.  I still plan on moving it to the `manageiq-api` repo, and will try and tackle that soon (aka:  this week).  Thanks!

djberg96:
@NickLaMuro Nice job, looking forward to it. :)

bdunne:
:turtle: Any progress on moving this over to the api repo?

NickLaMuro:
@bdunne nah... I think I will let it stew for another year...

NickLaMuro:
(kidding, I will take a look at taking care of this later today)

NickLaMuro:
And now this exists:

https://github.com/ManageIQ/manageiq-api/pull/767

So going to (_**FINALLY**_) close this PR.  Sorry for the delay everyone.

PR #319, opened 2018-07-17
Update documentation on host-based authentication

tadeboro:
Currently, developer setup docs fail to mention that tests also
require non-local access to PostgreSQL, which in turn causes those
tests to fail.

This commit remedies this situation by updating the PostgreSQL
configuration section of the docs with working pg_hba.conf file.

### DIFF
--- a/developer_setup.md
+++ b/developer_setup.md
@@ -40,8 +40,11 @@
 
   ```bash
   sudo postgresql-setup initdb
-  sudo grep -q '^local\s' /var/lib/pgsql/data/pg_hba.conf || echo "local all all trust" | sudo tee -a /var/lib/pgsql/data/pg_hba.conf
-  sudo sed -i.bak 's/\(^local\s*\w*\s*\w*\s*\)\(peer$\)/\1trust/' /var/lib/pgsql/data/pg_hba.conf
+  sudo cp /var/lib/pgsql/data/pg_hba.conf /var/lib/pgsql/data/pg_hba.conf.bak
+  cat <<EOF | sudo tee /var/lib/pgsql/data/pg_hba.conf
+  local   all             all                                     trust
+  host    all             root            127.0.0.1/32            trust
+  EOF
   sudo systemctl enable postgresql
   sudo systemctl start postgresql
   sudo -u postgres psql -c "CREATE ROLE root SUPERUSER LOGIN PASSWORD 'smartvm'"


bdunne:
@tadeboro Can you give some more detail about the error?  I think it's working fine for me without this change

tadeboro:
This is the relevant part of the rspec output:

```
Failures:

  1) MiqRegionRemote.with_remote_connection removes the temporary connection pool
     Failure/Error: conn = pool.connection

     PG::ConnectionBad:
       FATAL:  Ident authentication failed for user "root"
     # ./config/initializers/postgres_required_versions.rb:3:in `initialize'
     # ./app/models/miq_region_remote.rb:102:in `with_remote_connection'
     # ./spec/models/miq_region_remote_spec.rb:24:in `block (3 levels) in <top (required)>'
```

This specific test case tries to connect to database through `localhost` (all other tests connect using local unix socket) using `root` user. This, coupled with the fact that `pg_hba.conf`, as prepared by the upstream developerss, contains

    host    all             all             127.0.0.1/32            ident

line, makes this test fail.

tadeboro:
And why does CI works? This problem does not manifests itself on Travis, because PostgreSQL on that Ubuntu image is configured to allow those types of connections, but ManageIQ developer setup is missing those pieces.

Relevant Chef template: https://github.com/travis-ci/travis-cookbooks/blob/master/cookbooks/travis_postgresql/templates/ubuntu/pg_hba.conf.erb#L92

carbonin:
I think the instructions for `pg_hba.conf` and postgresql in general here are not intended to work for every possible use case that a manageiq developer could possibly encounter, but rather cover the easiest (and least intrusive) way to get the app up and running.

Generally most developers will not be running the entire test suite, and the ones who need to run the specs which make strange changes to the database configuration will know what changes they need to make to adjust.

I'm more comfortable with the current implementation of the guide as it leaves most of the initial `pg_hba.conf` in place rather than replacing the file entirely. Additionally I think some of the more security-minded folks would prefer not to allow any `host` lines at all in `pg_hba`.

tadeboro:
Maybe I am doing something weird, but when I setup my dev environment, first thing I do is run test suite to ensure that things are in working order. This way I can be sure that any errors I get when developing code are result of my own stupidity and are not baked into environment.

Following the guide in current form breaks this assumption, since there is a possibility that one of the tests will fail right from the start (and it indeed does, at least on Fedora 27/28).

Having no host lines in `pg_hba.conf` and have all tests pass are two incompatible things. What we can do is allow access only to test databases, but anything less than that will make test fail. But talking about security here is a bit pointless, since the first line makes database readily accessible to basically anyone with access to computer.

carbonin:
> Having no host lines in pg_hba.conf and have all tests pass are two incompatible things

Agree, and I'm on the side of not having all the tests pass OOTB because my opinion is that it is better to be less intrusive to the configuration than have tests pass which should be no concern for the vast majority of developers.

carbonin:
But someone with merge rights here may have a different opinion :smile:

tadeboro:
If changes to `pg_hba.conf` are too intrusive, we can just document what tests will fail if the root user cannot access test databases via `localhost` and instruct user to consult PostgreSQL manual for more information on `pg_hba.conf`.

martinpovolny:
@bdunne , @carbonin : merge or close, please. This is hanging here for looooooooong.

carbonin:
I would close it as I mentioned in my comment, but I don't have merge rights here.

gtanzillo:
I'm :+1 for closing per @carbonin and @martinpovolny discussion. However, I also do not have merge rights to this repo.

PR #322, opened 2018-07-27
Update UI test docs

tadeboro:
Instructions for running UI specs had a bad path in it, causing
`bin/update` in UI folder to fail with a message

    The git source https://github.com/ManageIQ/manageiq-gems-pending.git
    is not yet checked out. Please run `bundle install` before trying to
    start your application

This commit simply fixes this typo.

### DIFF
--- a/developer_setup/running_test_suites.md
+++ b/developer_setup/running_test_suites.md
@@ -128,8 +128,9 @@ file contains a block like this:
 
     override_gem "manageiq-ui-classic", path: "/home/u/miq/manageiq-ui-classic"
 
-Now you must run `bin/update` from the `/home/u/miq/manageiq` folder, followed
-by `bin/update` from the `/home/u/miq/manageiq-ui-classic` folder.
+Now you must run `bin/update` from the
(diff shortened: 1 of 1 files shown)


himdel:
I don't think it's a mistake in the docs.

You *are* supposed to use `manageiq`, not `manageiq-ui-classic/spec/manageiq`. Using a separate copy of `manageiq` is never mentioned in the docs (or is it?), and as far as I know, the original intent behind that was only to support easily running tests on travis.

The docs do mention having to create the symlink.

himdel:
Oh, also, bigger issue perhaps: the original text is lying anyway - you never have to run `bin/update` twice, there should be absolutely no difference between running `bin/update` from manageiq and from ui-classic (not when the symlink exists, that is).

And since line 124 of this file explicitly mentions creating the symlink...

himdel:
(The thing that needs to happen in ui-classic is `bundle update` so perhaps that was the original idea behind the double update.)

bdunne:
I think the documentation is correct the way it currently exists since it's using a symlink.  If it's the initial clone of manageiq, you need to run `bin/setup` there to set up the database.

PR #327, opened 2018-08-24
Don't require parens around add_reference

agrare:
Parentheses shouldn't be required around add_reference, see https://github.com/ManageIQ/manageiq-schema/pull/258#issuecomment-415752604

Follow up to https://github.com/ManageIQ/guides/pull/326


agrare:
@miq-bot assign @bdunne

agrare:
Thanks @bdunne

Issue #333, opened 2019-01-14
Can we use community styles for Percent Array literals?


NickLaMuro:
I have been frustrated with [our current override](https://github.com/ManageIQ/guides/pull/286) of [the defaults](https://github.com/rubocop-hq/rubocop/pull/4231) for a while now, and have not been completely silent about it in the past:

https://github.com/ManageIQ/manageiq-providers-amazon/pull/434#issuecomment-391060019
https://github.com/ManageIQ/manageiq-providers-amazon/pull/434#issuecomment-391752979

And I don't think [I am alone](https://github.com/ManageIQ/manageiq-providers-amazon/pull/434#issuecomment-391755128)...

I would like to petition we change this back to the community styles.

(I also willingly take on the work to update our code bases to use this style, as this would very much ease a long standing "erk-some" I have had with it)

Reference links
---------------

- `rubocop-hq` style guide discussion:    https://github.com/rubocop-hq/ruby-style-guide/issues/526
  - Case for change:  http://www.chrisrolle.com/blog/ruby-percentage-notations
  - Case against change (one I disagree with):  https://github.com/rubocop-hq/ruby-style-guide/issues/526#issuecomment-253888969
- PR overriding the defaults:  https://github.com/ManageIQ/guides/pull/286

Fryguy:
The only reason we had the overrides in the first place was because a) those were the original rubocop styles at the beginning until they changed it and b) our entire codebase was already that way.

Since there was never any good reason for it I am :+1: with going to the newer rubocop defaults.

> (I also willingly take on the work to update our code bases to use this style, as this would very much ease a long standing "erk-some" I have had with it)

Isn't it more or less just a `rubocop -a` command?

NickLaMuro:
> Isn't it more or less just a rubocop -a command?

Yeah, but I assume it is over multiple repos and with plenty of edge cases.  My main point is that I wasn't trying to off load work to some one else for a personal gripe that I had.

djberg96:
:+1:

bdunne:
I opened https://github.com/ManageIQ/guides/pull/286 to make rubocop properly enforce the cop that we had set in the config file.

My opinion: I don't like to use the `%w` type shortcuts at all.  I find `["a", "b"]` much easier than trying to remember whether `%w` gives me strings or symbols in it's effort to save a few characters.

NickLaMuro:
> I opened #286 to make rubocop properly enforce the cop that we had set in the config file.

[...]

PR #334, opened 2019-01-15
Disable Style/SymbolArray and Style/WordArray

NickLaMuro:
This change allows the use of either style of word arrays:

```ruby
["foo", "bar", "baz"] # both valid
%w[foo bar baz]
```

And symbol arrays:

```ruby
[:foo, :bar, :baz]    # both valid
%i[foo bar baz]
```

The style of symbol/word arrays will still be enforced, but will change in the future to match the defaults from `rubocop`.  Currently `%i()` and `%w()`, but will change to `%i[]` and `%w[]` in a future PR.

### DIFF
--- a/.rubocop_base.yml
+++ b/.rubocop_base.yml
@@ -240,6 +240,8 @@ Style/StringLiterals:
   Enabled: false
 Style/StringLiteralsInInterpolation:
   Enabled: false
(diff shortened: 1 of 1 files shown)


Fryguy:
Merging after the discussion in #333 and the talk.manageiq.org poll

PR #344, opened 2019-03-14
Guide for installing a stable node on mac

kbrock:
Fixes an error while running `bin/update`.
Under the covers it was failing from task `rake update:ui`

```
ERROR in ./node_modules/spin.js/spin.css (
./node_modules/css-loader??ref--8-1!
./node_modules/postcss-loader/lib??ref--8-2!
./node_modules/resolve-url-loader!
./node_modules/sass-loader/lib/loader.js??ref--8-4!
./node_modules/spin.js/spin.css)

Module build failed (from ./node_modules/sass-loader/lib/loader.js):
Error: Node Sass does not yet support your current environment:
OS X 64-bit with Unsupported runtime (67)

For more information on which environments are supported please see:
https://github.com/sass/node-sass/releases/tag/v4.9.4
```

Apparently odd numbers for node are unstable, so I needed to downgrade from v9 to v8

/cc @himdel

### DIFF
--- a/developer_setup.md
+++ b/developer_setup.md
@@ -116,7 +116,7 @@
   brew install git
   brew install pkg-config
   brew install memcached
-  brew install postgresql
+  brew install postgresql@9.5
   brew install cmake
   brew install node
   brew install yarn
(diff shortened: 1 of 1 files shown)


himdel:
LGTM, brew does tend to install inconvenient versions sometimes :+1:

PR #352, opened 2019-05-14
Update documentation for new plugin generators

Fryguy:
@jrafanie Please review.


plugins/generator.md
```diff
@@ -0,0 +1,91 @@
+## Generators
+
+### Plugin generator
+
+Use the plugin generator to create a ManageIQ plugin.
+
+For example, to create a plugin called ManageIQ::MyPlugin, do the following
```
jrafanie: Do we have an explanation of when to use the ManageIQ namespace or not?  Maybe it doesn't belong here but that was my first thought...

plugins/generator.md
```diff
@@ -0,0 +1,91 @@
...
+the creation progress emitted to the console:
+
+```text
+create
+create  .codeclimate.yml
+...
+```
+
+By default, the plugin will be created in the `plugins` directory, but you can
+specify an alternate directory with `--path <path>`.
```
jrafanie: We might want to articulate how the path is build normally for namespacing and if that's important to follow even if it's in a different `--path`.

plugins/generator.md
```diff
@@ -0,0 +1,91 @@
...
+cd manageiq
+bundle exec rails generate manageiq:provider ManageIQ::Providers::AwesomeCloud
+```
+
+Much like the plugin generator, you can pass `--path` to change the path.
+Additionally, you also have the following options
+
+```text
+[--vcr], [--no-vcr]      # Enable VCR cassettes (Default: --no-vcr)
+[--dummy], [--no-dummy]  # Generate dummy implementations (Default: --no-dummy)
```
jrafanie: What's a dummy implementation?

plugins/generator.md
Fryguy: There is no need to use the ManageIQ namespace, so I didn't mention it.  I'm not sure how I would express that.

plugins/generator.md
Fryguy: We don't build the path...it's the directory into which the plugin will be created.

plugins/generator.md
Fryguy: It's described below in the output ... I'll tweak this to make that clearer.

Fryguy:
@jrafanie, Ok, I shuffled around the bottom section to make it clearer and to make the description live with the line item.

PR #354, opened 2019-05-16
Change default preferred exception name to 'err'

djberg96:
As of rubocop 0.5.3, they added the rule `Naming/RescuedExceptionsVariableName`. This checks the variable name assigned to exceptions when you do something like this:

`rescue StandardError => err # <== I'm talking about 'err' here`

It's a nice idea in theory. Unfortunately, they chose a default preferred value of "e", which means that any other variable name will generate a rubocop warning about the exception name.

I think violates the whole "meaningful variable names" concept. I think "err" hits the sweet spot between meaningful and concise, so I'm suggesting this tweak.

Edit: in fact, it violates their *own* default variable naming enforcement for the also-new `Naming/UncommunicativeMethodParamName` option, which defaults to a minimum of 3 characters.

If you said "either/or" I'll count one vote for each.

Vote tally:

* Use `err`: 3
* Use `error`: 2
* Just stick with `e`: 1
* Disable the cop: 4

Update: Now it's just disabled since that seems to have the edge in the voting.

### DIFF
--- a/.rubocop_base.yml
+++ b/.rubocop_base.yml
@@ -38,6 +38,8 @@ Metrics/ParameterLists:
   Enabled: false
 Metrics/PerceivedComplexity:
   Enabled: false
+Naming/RescuedExceptionsVariableName:
+  Enabled: false
 Performance/Casecmp:
   Enabled: false
 Rails:


bdunne:
Looks like both `e` and `err` are very common in the core code base.  I agree and don't really like just `e`, so I vote for changing it as proposed or disable the cop.

NickLaMuro:
I think I share @bdunne 's opinion on this one.  Could go for either `err` or `disabled`.

I guess I would prefer `disabled` since I usually don't mind writing out `error`, but maybe I am just silly.

bdunne:
I also prefer `error`, but there are significantly fewer of those in the code base

martinpovolny:
I do not like `err`. I like `error` or just `e`.

d-m-u:
+1 @bdunne

Fryguy:
I vote for `disabled`.

@djberg96 Can you keep a running tally in the OP post?

djberg96:
@Fryguy I'll try, people are suggesting multiple options, though.

jrafanie:
I prefer disabling the cop.  It seems meaningless, especially if `e` is picked up by the `Naming/UncommunicativeMethodParamName`

djberg96:
@jrafanie It doesn't appear to actually violate it in practice, I guess because of how it's parsed. But I'll note your vote.

djberg96:
@bdunne @Fryguy @jrafanie @NickLaMuro @martinpovolny @d-m-u Seems most people were indifferent to the cop and generally favored disabling it, so I've updated the PR to just disable it.

In the future I hope they add a whitelist option for this, but that doesn't seem to be an option currently.

jrafanie:
@djberg96 this might be the one time I'll recommend squashing the commits since the first one is reversed in the second one... once we've got consensus  🤣

djberg96:
@jrafanie ok, squished.

PR #356, opened 2019-06-18
update postgresql-setup call

mshriver:
Fed30 default is pg11, installing from Postgres GDG doesn't have postgresql-setup in PATH

### DIFF
--- a/developer_setup.md
+++ b/developer_setup.md
@@ -42,9 +42,13 @@
 
   ```bash
   sudo postgresql-setup initdb
+  # use below if postgresql-setup is not in PATH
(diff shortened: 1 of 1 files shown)


chessbyte:
@carbonin @Fryguy @bdunne can you take a look at this PR

bdunne:
The only issue I see is that we're again going down the path of documenting how to configure Postgres from multiple sources.  But if we had to pick a single source, I think it would be "installing from Postgres GDG", so let's rebase and move forward with this.

chessbyte:
@mshriver Please re-base and re-open, if you want to move this PR forward

PR #363, opened 2019-07-29
Rails/HttpStatus - disable cop

himdel:
Proposing to disable the [Rails/HttpStatus](https://rubydoc.info/gems/rubocop/0.69.0/RuboCop/Cop/Rails/HttpStatus) cop

The current setting for the cop is to prefer symbols, which means:

```
# "bad"
render :foo, status: 200
render json: { foo: 'bar' }, status: 200
render plain: 'foo/bar', status: 304
redirect_to root_url, status: 301

# good
render :foo, status: :ok
render json: { foo: 'bar' }, status: :ok
render plain: 'foo/bar', status: :not_modified
redirect_to root_url, status: :moved_permanently
```

I consider the numeric codes more readable and clearer,
but would not go as far as forcing people to use them.

(Translation table between actual status codes and rails symbols: https://guides.rubyonrails.org/layouts_and_rendering.html#the-status-option)

=> just disabling the cop in this PR

Please vote on what you prefer with :+1: or :-1: on this comment. (:+1: means disable the cop, :-1: means `:status => :ok` is the only right way)

### DIFF
--- a/.rubocop_base.yml
+++ b/.rubocop_base.yml
@@ -47,7 +47,8 @@ Rails:
 Rails/FindEach:
   Enabled: false
 Rails/HttpStatus:
-  Enabled: false
+  Enabled: true
+  EnforcedStyle: numeric
 Rails/ReadWriteAttribute:
   AutoCorrect: false
 Style/BracesAroundHashParameters:
--- a/.rubocop_base.yml
+++ b/.rubocop_base.yml
@@ -46,6 +46,8 @@ Rails:
   Enabled: true
 Rails/FindEach:
   Enabled: false
+Rails/HttpStatus:
+  Enabled: false
 Rails/ReadWriteAttribute:
   AutoCorrect: false
 Style/BracesAroundHashParameters:


himdel:
Alternative proposal in this comment:

Don't disable the cop, switch the preferred style to numeric, making `:status => :ok` bad, and `:status => 200` the only right way.

Please vote on what you prefer with +1 or -1 on this comment.
(+1 means switch to `numeric`, :-1: means don't switch to numeric)

NickLaMuro:
@himdel honestly with the addendum comment, this is pretty confusing as people have voted, since some have voted in both comments and the original post.  However, I see those two options as mutually exclusive and it is uncertain what vote counts for what.  Personally, I think polling should be done in a single spot (though I guess I am indifferent if you use talk.manageiq.org or something else), but don't split the decision between two comments.

Example:  What I did here for array literals:

https://github.com/ManageIQ/guides/issues/333

* * *

I personally voted for going with enforcing `numeric` only over using the `symbol` form as `:ok` and friends is pretty vague.  In general, I would prefer that we are consistent throughout our applications instead of "whatever the developer feels like at the time" (which then begs the question if `rubocop` is serving any purpose at that point...), so I think the original suggestion doesn't make sense.

But the above is also why your voting system is unclear as it is hard to tell if what you have is:

> "choose between `a`, `b`, or `c`"

or 

> "choose `a` or `b`, and if `a`, pick `a1` or `a2`"

or something else entirely...

himdel:
..Yeah, sorry about that, but I think it doesn't have to be mutually exclusive:

parse the root comment as "do I want rubocop to complain when I use numbers?",
and the addendum as "do I want rubocop to complain when I use symbols?"

himdel:
Either way, looking at the current code:

manageiq-ui-classic:
symbol - 2
number - 10

manageiq-api:
symbol - 0
number - 1

Fryguy:
@himdel If you haven't already can you create a talk article linking back to this issue?  That way devs that don't follow guides can at least find out about this vote.

himdel:
Good idea, thanks :) 

http://talk.manageiq.org/t/vote-rubocop-rails-httpstatus-disable-cop-or-switch-to-prefer-numeric/4486

Fryguy:
Looks like there is agreement to enforce numeric, so I'm going to merge this.

PR #369, opened 2019-08-13
Revert #137 and enforce no spaces inside hash braces

bdunne:
As a reviewer, I'd prefer to re-enable this cop for a couple of reasons:
- With it disabled, it is harder to differentiate between a hash and an inline block
- With the cop disabled, it doesn't complain about inconsistencies in leading and trailing spaces like `{:a => 1 }` or `{ :a => 1}`

This change only affects hashes, not in-line blocks and enforces no leading or trailing spaces inside the braces.

Please vote :+1: if you approve of the change or :-1: if you prefer to leave it disabled.  I think there was a lot of unrelated discussion about Arrays, Blocks and String interpolation (which are not controlled by this cop) in #137 which led to confusion.

### DIFF
--- a/.rubocop_base.yml
+++ b/.rubocop_base.yml
@@ -17,7 +17,7 @@ Layout/ExtraSpacing:
 Layout/SpaceBeforeFirstArg:
   Enabled: false
 Layout/SpaceInsideHashLiteralBraces:
-  Enabled: false
+  EnforcedStyle: no_space
 Metrics/AbcSize:
   Enabled: false
 Metrics/BlockLength:


Fryguy:
Didn't see this vote...but you should probably promote it on gitter and talk.manageiq.org if you want more people to vote.

.rubocop_base.yml
```diff
@@ -17,7 +17,7 @@ Layout/ExtraSpacing:
 Layout/SpaceBeforeFirstArg:
   Enabled: false
 Layout/SpaceInsideHashLiteralBraces:
-  Enabled: false
+  EnforcedStyle: no_space
```
Fryguy: Is the default no_space?  If so, you can just remove this cop from the file altogether.

bdunne: No, the default is `space`

Fryguy:
I'm confused a bit about what the "before" and "after" are after the revert...can you show examples?

bdunne:
> I'm confused a bit about what the "before" and "after" are after the revert...can you show examples?

Currently the cop is disabled so anything goes, spaces, no spaces, and inconsistencies.
With this change...

```ruby
Good: {:a => 1}

Bad: { :a => 1 }
Bad: { :a => 1}
Bad: {:a => 1 }
```

bdunne:
> Didn't see this vote...but you should probably promote it on gitter and talk.manageiq.org if you want more people to vote.

http://talk.manageiq.org/t/rubocop-setting-for-spaces-inside-hash-literal-braces/4551

Fryguy:
I'm curious why in #137 we went down the path of this settings affecting blocks and interpolation when it doesn't.  Since I prefer Hashes and interpolation without spaces, but blocks with spaces, I voted up that other PR because my choice was inconsistent.  If you're saying it really only affects Hashes, then I'm :+1: on this PR.

bdunne:
> I'm curious why in #137 we went down the path of this settings affecting blocks and interpolation when it doesn't. Since I prefer Hashes and interpolation without spaces, but blocks with spaces, I voted up that other PR because my choice was inconsistent. If you're saying it really only affects Hashes, then I'm  on this PR.

Yes, this only affects hashes. https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Layout/SpaceInsideHashLiteralBraces and there are other cops for the rest of those.
I think too many things were being discussed in #137 so I'm trying to address them individually.

himdel:
How would this change affect multi-line hashes?

I'm all :+1: forcing just `{:foo => 5}` but not if it means we can't use

```
{
  :foo => 5,
  :bar => 6,
}
```

himdel:
Tested.. 

```
ok1 = {:foo => 5, :bar => 6}
ok2 = {
  :foo => 5,
  :bar => 6,
}
ok3 = {:foo => 5,
       :bar => 6}

bad1 = { :foo => 5 }
```

```
foo.rb:12:9: C: Layout/SpaceInsideHashLiteralBraces: Space inside { detected.
bad1 = { :foo => 5 }
        ^
foo.rb:12:19: C: Layout/SpaceInsideHashLiteralBraces: Space inside } detected.
bad1 = { :foo => 5 }
```

:+1: from me :)

slemrmartin:
> It's been a few months and we've had so few votes. There is one sure fire way to get feedback on this: Merge and see what people say.

@jrafanie Really? Push some voting somewhere on manageiq's pages, then change rubocop rules we are using for c.r.c.'s services and *see what people say* ??? I'm saying I really don't like it. 

Okay, I accepted #137 rule, fine. But I don't want to changing styling everytime when 2-3 guys suddenly decide that everyone has to change style of coding 

BIG :-1: For this PR!

bdunne:
@slemrmartin Much of the code base already conforms to this rule since it used to be in effect.  I think there was a lot of confusion around #137 and it wound up being disabled for the wrong reason.  This PR re-enabled it with the same settings that it used to have, so I don't think it's a huge style change.

> But I don't want to changing styling everytime when 2-3 guys suddenly decide that everyone has to change style of coding

This was not a closed discussion in any way.  It was open for anyone to join in and vote or comment for almost 5 months.  It is typically the same few people (mostly reviewers) making PRs to our style guide because it helps reviewers most.  Anything that rubocop can do for us to make PR review easier results in PRs getting merged faster.

> Really? Push some voting somewhere on manageiq's pages, then change rubocop rules we are using for c.r.c.'s services and see what people say ??? I'm saying I really don't like it.

The process for changing the style is typically:
- open a PR here
- add the `vote` label
- mention it in Gitter
- and usually post it to talk.manageiq.org as well
- wait at least 2 weeks for feedback/votes

We will make exceptions and merge things sooner if needed.  For example:
- if a rubocop upgrade introduces a new cop that severely conflicts with the code base we will quickly merge a PR to disable it.
- if a rubocop upgrade switches style enforcement we'll quickly merge a PR that sets it back to the previous style.

Do you have any other suggestions to get the word out there in the future?

jrafanie:
@slemrmartin Please let us know how we missed getting your feedback earlier.  I couldn't tell what you disliked about this change.  Regarding merging with little votes:  there is very little we can do if we post this on talk.manageiq.org, discuss it in gitter, elsewhere and yet get very little feedback.  If there is more that we can do to get more feedback other than what @bdunne outlined, please let us know.  

Unfortunately, though, if we still only get feedback from the same group of people even though we try to get other feedback, there is very little we can do other than to merge it and get feedback after the fact.

These guidelines are not set in stone and if people want something reverted/changed with a different PR that includes feedback on why, we can certainly do that.

slemrmartin:
@jrafanie we discussed it and we could split rubocop rules for ManageIQ and RedHatInsights organizations. 
Then these rules can be announced on talk.manageiq.org and RedHatInsights's rules (I suggest) for example on some mailing list (and somewhere else related to this org if needed).

PR #372, opened 2019-08-29
Don't set pull.rebase to true by default

subpop:
Rebasing by default requires knowing how to resolve conflicts that arise from rebasing. I know that rebasing can keep commit history clean(er) by keeping out merge commits, but it can make for some complicated working trees that are hard to resolve cleanly.

For new developers, I think it's best if they don't have to learn rebasing techniques at the same time as learning a new project & code base.


subpop:
I'm second-guessing myself, so I'm going to drop this for now.

PR #376, opened 2019-09-17
Add methods to be ignored by rubocop

astrozzc:
These methods are  usually treated as a DSL and should be ignored
by Style/MethodCallWithArgsParentheses.

See comments in https://github.com/ManageIQ/manageiq-automation_engine/pull/369#discussion_r324851649
for more details.

### DIFF
--- a/.rubocop_base.yml
+++ b/.rubocop_base.yml
@@ -93,6 +93,9 @@ Style/MethodCallWithArgsParentheses:
   Enabled: true
   IgnoredMethods:
     # Ruby
+    - add_dependency
+    - add_development_dependency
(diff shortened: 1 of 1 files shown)


.rubocop_base.yml
```diff
@@ -197,7 +197,10 @@ Style/MethodCallWithArgsParentheses:
     - die
     - opt
     # MIQ
+    - add_dependency
```
bdunne: These should probably be under `# Ruby` above or a new `# Rubygems` since the methods are not defined in the ManageIQ code base.

astrozzc: Good point! Updated

PR #382, opened 2019-11-13
Some parameter names are short but meaningful

gekorob:
Parameters like 'vm' and 'dc' are quite meaningful and often used in
providers so I propose to make an exception to the rule of minimum three
characters length for those.


agrare:
:+1: these are use throughout core

gekorob:
Thx a lot guys

PR #395, opened 2020-02-07
Create triage_process.md

mfeifer:
Make triage process and meeting times available to the community.


mfeifer:
@Fryguy trying to decide if stale should be in a different document.
also, I need to add info for ui/api triage

PR #397, opened 2020-02-18
Create triage_process.md

mfeifer:
Includes instructions and links to each category of triage

### DIFF
--- a/triage_process.md
+++ b/triage_process.md
@@ -1,0 +1,54 @@
+# Issue and PR Lifecycle and Triage Process
+
+The community holds meetings regularly to triage issues and pull requests across the ManageIQ organization’s repositories. Attendees include code owners and mergers for the repositories, but all are invited to attend.*This is not a meeting to assign when work will be done.*
+
+Triage meetings are divided into two groups General and API/UI.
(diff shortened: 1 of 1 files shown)


triage_process.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Issue Triage
+* Triage unassigned issues
+  * Assign the issue.  There are three possible scenarios.
+    * Assign someone to at least get in on the radar.  This *does not* mean that the assignee is expected to start work on the issue.
+    * Apply the `help wanted` label.  This means that while the issue is valid, the team does not have plans to work on it in the near future.
+    * Do not assign to anyone and revisit the issue.  In other words, we want to be sure that the issue does not get ignored in perpetuity.
+  * Close invalid issues with a comment and/or label as to why it has been closed.
+* Review issues with assignee, but no scope label.
+  * Apply [scope labels](https://www.manageiq.org/docs/guides/labels).
```
Fryguy: Instead of a straight URL you can use a relative path to the labels document.  I believe this should translate properly on the website.  So, something like

[scope labels](labels.md)

Fryguy:
I'm wondering if we should have a "process" subdirectory where we put this and the merger do's and don'ts, but we can move later, so I'm merging.

PR #401, opened 2020-02-28
Create backport_process.md

mfeifer:
Document to publicize new backport process.

### DIFF
--- a/backport_process.md
+++ b/backport_process.md
@@ -1,0 +1,11 @@
+# Backport Process
+
+Beginning with the Jansa release of ManageIQ, we will be using the following backport process.  Here are the details!
+
+* We are adding a new label `release_to_backport_to/yes?` 
+* Anyone can request a backport by typing `@miq-bot add_label release_to_backport_to/yes?` in a comment in the pull request.
+* During the [triage meetings](triage_process.md), the pull request will be evaluated for backportability.  
+  * The label will be set to `release_to_backport_to/no` if the backport should be denied for any reason.
+  * The label will be set to `release_to_backport_to/yes` if the pull request should be backported.
+* The `release_to_backport_to/no` and `release_to_backport_to/yes` labels can only be set by those with special access.
+* When the pull request has been backported, the label will flip to `release_to_backport_to/backported`.


backportprocess.md
```diff
@@ -0,0 +1,12 @@
+# Backport Process
+
+Beginning with the Jansa release of ManageIQ, we will be using the following backport process.  Here are the details!
+
+* We are adding  a new label `release_to_backport_to/yes?` Anyone can set this label to request that a backport.
```
Fryguy: 2 spaces

backportprocess.md
```diff
@@ -0,0 +1,12 @@
+# Backport Process
```
Fryguy: Can we name the file `backport_process.md`, which is a common pattern?

backportprocess.md
Fryguy: > Anyone can set this label to request that a backport. 

Sort of?  They can't actually apply it directly.

We should say something like, "If you'd like to apply this label, you can ask miq-bot with `@miq-bot add_label release_to_backport_to/yes?`".  or something to that effect.

backportprocess.md
```diff
@@ -0,0 +1,12 @@
...
+
+Beginning with the Jansa release of ManageIQ, we will be using the following backport process.  Here are the details!
+
+* We are adding  a new label `release_to_backport_to/yes?` Anyone can set this label to request that a backport. 
+* During the [triage meetings](triage_process.md), the pull request will be evaluated for backportability.  These labels can only be set by those with special access.
+  * The label will be set to `release_to_backport_to/no` if the backport will break that release.
+  * The label will be set to `release_to_backport_to/yes` if the pull request should be backported.
+* When the pull request has been backported, the label will flip to `release_to_backport_to/backported`
+
```
Fryguy: Minor, but too many blank lines at end.

backport_process.md
```diff
@@ -0,0 +1,12 @@
+# Backport Process
+
+Beginning with the Jansa release of ManageIQ, we will be using the following backport process.  Here are the details!
+
+* We are adding a new label `release_to_backport_to/yes?` Anyone can set this label to request a backport. 
+* During the [triage meetings](triage_process.md), the pull request will be evaluated for backportability.  These labels can only be set by those with special access.
```
chessbyte: "These labels can only be set ..." text seems out of place - because the labels will be described below.   Instead, suggest moving this sentence as the 3rd bullet.
Separately, people can set `release_to_backport_to/no`, IIRC, to basically say that this should **not** be backported because reasons.  /cc @Fryguy

backportprocess.md
chessbyte: agree with the text to request via miq-bot

backport_process.md
```diff
@@ -0,0 +1,10 @@
+# Backport Process
+
+Beginning with the Jansa release of ManageIQ, we will be using the following backport process.  Here are the details!
+
+* We are adding a new label `release_to_backport_to/yes?` Anyone can set this label to request a backport.
```
chessbyte: How about enhancing "Anyone can set this label to request a backport." to "Anyone can set this label to request a backport via `@miq-bot add_label release_to_backport_to/yes?`" or something to that effect.

Fryguy: Doh...I think something else changed on the line and I marked the other one as resolved seeing it as Outdate.

PR #407, opened 2020-03-05
Add script to generate the sprint boundaries

Fryguy:
@bdunne  Please review.  This script picked up the fact that the existing document has an extra line.


Fryguy:
@bdunne Note that I extracted this from manageiq-release...I expect to have a deletion PR on the other side since we don't use that anymore.

PR #416, opened 2020-04-08
Add instructions for adding CodeReady Containers

agrare:
Add instructions for using CodeReady Containers as a local OpenShift 4 provider.

Support for OpenShift 4/CRC was added in https://github.com/ManageIQ/manageiq-providers-openshift/pull/164

### DIFF
--- a/providers/openshift.md
+++ b/providers/openshift.md
@@ -67,7 +67,7 @@ $ ansible-playbook playbooks/byo/config.yml -i path/to/inventory/file
 
 For a simple single host setup for working with the openshift provider locally,
 you can use [`minishift`](https://github.com/minishift/minishift) to run a
-local openshift instance.
+local openshift 3.x instance.
 
 
 #### Requirements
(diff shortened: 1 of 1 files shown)


providers/openshift.md
```diff
@@ -136,14 +136,88 @@ regardless of the hypervisor).
...
+    ? Image pull secret [? for help]
+    ```
+
+4. Login to the cluster with `oc`, the command will be printed when `crc start` finishes or you can call `crc console --credentials`
+    ```console
+    crc console --credentials
+    oc login -u kubeadmin -p KUBEADMING_PASSWORD https://api.crc.testing:6443
+    ```
+
+5. Now you can setup the management-infra/management-admin service account for use with ManageIQ
```
agrare, line 187: This section is basically https://gist.github.com/NickLaMuro/e2fac8be87ea0e9f429b6f5d75e02176#file-manageiq-addon without the hawkular bits

Thinking we should move this into the openshift repo instead of in a gist

Fryguy: This is fine for now

PR #421, opened 2020-05-05
Updating rubocop config to work with rubocop-0.82.0

NickLaMuro:
This is part of the bot effort here:

https://github.com/ManageIQ/miq_bot/issues/496


Fryguy:
Closing now that we have #443 and https://github.com/ManageIQ/manageiq-style

PR #432, opened 2020-07-08
Update i18n.md to replace Zanata with Transifex

chessbyte:
@mzazrivec Please review

### DIFF
--- a/i18n.md
+++ b/i18n.md
@@ -13,11 +13,9 @@
 * [Translating ManageIQ](#translating-manageiq)
   * [Translator notes](#translator-notes)
 * [Updating translations](#updating-translations-developers)
-  * [Zanata](#zanata)
+  * [Transifex](#transifex)
   * [ManageIQ](#manageiq-the-main-project)
   * [ManageIQ Self Service UI](#manageiq-self-service-ui)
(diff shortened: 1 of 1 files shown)


i18n.md
```diff
@@ -390,57 +387,3 @@ $ gulp available-languages
...
-
-* Commit and push the new catalogs into git:
-
-```sh
-$ git commit locale/
-$ git push
-```
-
-
-## Internationalization in rails engines (plugins)
```
Fryguy, line 424: I think this part is still available, but optional.

chessbyte: @Fryguy I was under the impression that the translations would be happening in https://github.com/ManageIQ/manageiq repo once the plugin is included in the ManageIQ build, and not in individual plugin repos.

mzazrivec: Yes, I made the change to include strings from all enabled plugins in one ManageIQ resource. And yes, it is also still possible to have a separate catalog for your ManageIQ plugin, but we did that only for plugins that are not enabled & shipped with ManageIQ by default (cloud services & migration analytics). So I'd say it's okay to remove this part from ManageIQ guidelines.

PR #436, opened 2020-08-11
Minor fixes to reflect current code setup for Internationalization

chessbyte:


chessbyte:
@mzazrivec please review

i18n.md
```diff
@@ -307,14 +307,14 @@ Now, commit and push changed files into git (branch, pull request, etc.). Make s
 config/model_attributes.rb
 config/dictionary_strings.rb
 config/yaml_strings.rb
-config/locales/manageiq.pot
-config/locales/*/*.po
+locale/manageiq.pot
+locale/*/*.po
```
mzazrivec, line 311: This change is OK, but may I also suggest to remove the other 3 files as well:
```
config/model_attributes.rb
config/dictionary_strings.rb
config/yaml_strings.rb
```

These 3 files do not need to be commited into git, so we can drop them from the docs as well.

chessbyte: Done!

@mzazrivec Separate question - why should we be checking in locale/*/*.po files in this step?  What changes are happening to them?

mzazrivec: @chessbyte Come to think of it, the `locale/*/*.po` files do not need to be commited in this step. The `.po` files only need to be updated when new translations (for example from Transifex) arrive.

The `.po` files do get updated at this point though: the strings extraction updates the main `.pot` file and at the same times updates the `.po` files so that they do not contain strings that went away and do contain strings that are new.
Sometimes though I've noticed that these automatic `.po` updates are not 100% reliable, especially if the library attemts
to do fuzzy translations.

If the updated `.po` files won't be commited to git in this step -- nothing happens really. The gettext library is able to handle absent / extraneous strings in `.po` files just fine.

chessbyte: perfect - so I will remove those from this step as well and just mention the .pot file.

mzazrivec:
Approved, cannot merge though (no merge rights).

PR #437, opened 2020-08-29
`tx push --source` uses the `.tx/config` file to determine the message catalog file name

chessbyte:
As an example, [this](https://github.com/ManageIQ/manageiq/blob/master/.tx/config#L6) is the message catalog in the [manageiq](https://github.com/ManageIQ/manageiq) repo.

### DIFF
--- a/i18n.md
+++ b/i18n.md
@@ -311,7 +311,7 @@ locale/manageiq.pot
 
 ```sh
 $ cd locale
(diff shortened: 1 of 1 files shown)


chessbyte:
@mzazrivec please review

mzazrivec (approved):

PR #439, opened 2020-09-17
Add Backport direct PRs link to triage

Fryguy:


Fryguy:
cc @simaishi

bin/triage.yml
```diff
@@ -35,6 +35,13 @@ params:
   - is:merged
   - is:pr
   - label:jansa/yes?
+  backport_direct_prs: &backport_direct_prs
+  - base:jansa
+  - is:open
+  - is:pr
+  - -label:jansa/backported
```
simaishi, line 42: If it's open, it should never have `/backported`, but it won't hurt having it.

Fryguy: Yeah, I'll keep it for now as a "just in case".  I left conflict out on purpose cause we may want that one to show up in the triage.

Fryguy: Actually, I missed the nuance that _open_ shouldn't have `/backported`, so I removed it.

gtanzillo (approved):

PR #442, opened 2020-10-14
Update the minimum NodeJS version to 12

skateman:
cc @himdel please verify

### DIFF
--- a/developer_setup.md
+++ b/developer_setup.md
@@ -7,7 +7,7 @@
 | Ruby       | 2.5.x       |
 | Rails      | 5.1.x       |
 | Bundler    | 1.15.x      |
-| NodeJS     | 10.16.x     |
+| NodeJS     | 12.x.x      |
 | Python     | 2.7.x       |
 | PostgresQL | 10.x        |
 


himdel:
LGTM :+1: (everything should still work with v10, but we're not building or testing with it anymore)

skateman:
Nope, we don't allow it below 12 :confused:  https://github.com/ManageIQ/manageiq-ui-classic/blob/master/package.json#L173 that's

himdel:
It's more of a recommendation really :) but yeah, agreed, this is :+1:

jrafanie (approved):
LGTM, can you verify the other branches have the right version?

himdel:
Good point, according to package.json:

ivanchuk: >8 <12
jansa: >=12
master: >=12

Is there a guides equivalent of jansa/yes?

jrafanie:
> Good point, according to package.json:
> 
> ivanchuk: >8 <12
> jansa: >=12
> master: >=12
> 
> Is there a guides equivalent of jansa/yes?

It doesn't look like we have release branches for this repo.  Does it make sense to have branches with the setup for those branches or put all of the setup information on master and point out what's different in terms of software versions (ruby/postgresql/node/etc.). Maybe we can point to [cfme-versions](https://github.com/RedHatCloudForms/cfme-versions), even though node doesn't look like it's in there yet.

If you've ever switched branches and not been aware of where to look to figure out how to get the javascript dependencies correct, it's painful.  I imagine it's probably the same for ruby/postgresql if you're also not sure where to look.  I'd prefer we delegate version information to one place to maintain, such as `cfme-versions` or something similar.

cc @Fryguy @chessbyte

himdel:
(Might be easier to ivanchuk/yes https://github.com/ManageIQ/manageiq-ui-classic/pull/6008 and solve it that way..)

Fryguy:
Developer setup is about getting new developers setup for development on master and so should only be master oriented.  We don't, for example, tell people to install older rubies just because we used older rubies in the past.  This repo is not versioned.  If something needs versioning (and in this case I don't think it does), we can use manageiq documentation, however that's generally just for end user documentation.

Imo, what we _should_ do is on older branches of manageiq-ui-classic, we should raise a clear error when the expected version of a dependency is not met.  We do this for Ruby and PG on the MiQ core side, so that when devs go to old branches they aren't caught off guard.

Fryguy:
@himdel This does raise a question I had about package.json.  It specifies dependencies for things like node and yarn, but does it enforce it?  I would have expected something to verify that value and raise a clear error.

himdel:
If you `bin/update` (or really just `yarn`), it does...

```
error manageiq-ui-classic@1.0.0: The engine "node" is incompatible with this module. Expected version ">= 16.0.0". Got "14.6.0"
error manageiq-ui-classic@1.0.0: The engine "yarn" is incompatible with this module. Expected version ">= 2.0.0". Got "1.22.5"
```

PR #458, opened 2021-04-13
Various dev setup updates

Fryguy:
Reviewing these steps with some new developers, I found some things lacking or unclear, so this PR fixes those.

@chessbyte Please review.  (easier to review with whitespace diffs off)

### DIFF
--- a/developer_setup.md
+++ b/developer_setup.md
@@ -13,132 +13,127 @@
 
 ## Prerequisites
 
-If running CentOS 8
+### System packages
 
-```bash
-sudo yum config-manager --set-enabled PowerTools
-sudo dnf copr enable manageiq/ManageIQ-Master
-sudo yum install epel-release
-```
+1. Only if running CentOS 8
 
-In order to compile Ruby, install the header files for OpenSSL, readline and zlib.
+   ```bash
+   sudo yum config-manager --set-enabled PowerTools
+   sudo dnf copr enable manageiq/ManageIQ-Master
+   sudo yum install epel-release
+   ```
 
-|      |     |
-| ---- | --- |
-| dnf  | `sudo dnf -y install openssl-devel readline-devel zlib-devel` |
-| yum  | `sudo yum -y install openssl-devel readline-devel zlib-devel` |
-| apt  | `sudo apt -y install libssl-dev libreadline-dev zlib1g-dev` |
-| brew | `brew install openssl` |
+2. In order to compile Ruby, install the header files for OpenSSL, readline and zlib.
 
+   |      |     |
+   | ---- | --- |
+   | dnf  | `sudo dnf -y install openssl-devel readline-devel zlib-devel` |
+   | yum  | `sudo yum -y install openssl-devel readline-devel zlib-devel` |
+   | apt  | `sudo apt -y install libssl-dev libreadline-dev zlib1g-dev` |
+   | brew | `brew install openssl` |
 
-In order to compile Ruby, native Gems and native NodeJS modules, other libraries are required.
+3. In addition, in order to compile Ruby, native Gems and native NodeJS modules, other libraries are required.
 
-|      |     |
-| ---- | --- |
-| dnf  | `sudo dnf -y install @c-development libffi-devel postgresql-devel libxml2-devel libcurl-devel cmake python` |
-| yum  | `sudo yum -y install @development libffi-devel postgresql-devel libxml2-devel libcurl-devel cmake python` |
-| apt  | `sudo apt -y install build-essential libffi-dev libpq-dev libxml2-dev libcurl4-openssl-dev cmake python` |
-| brew | `brew install cmake` |
+   |      |     |
+   | ---- | --- |
+   | dnf  | `sudo dnf -y install @c-development libffi-devel postgresql-devel libxml2-devel libcurl-devel cmake python` |
+   | yum  | `sudo yum -y install @development libffi-devel postgresql-devel libxml2-devel libcurl-devel cmake python` |
+   | apt  | `sudo apt -y install build-essential libffi-dev libpq-dev libxml2-dev libcurl4-openssl-dev cmake python` |
+   | brew | `brew install cmake` |
 
-### Prerequisite services
+### Services
 
 ManageIQ requires a memcached instance for session caching and a PostgreSQL database for persistent data storage.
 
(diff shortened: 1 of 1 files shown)


developer_setup.md
```diff
@@ -13,145 +13,140 @@
...
-| brew       | `brew install postgresql` |
-| containers | N/A |
-
-###### NOTE
+   |            |     |
+   | ---------- | --- |
+   | dnf        | `sudo dnf -y install postgresql-server` |
+   | yum        | `sudo yum -y install postgresql-server` |
+   | apt        | `sudo apt -y install postgresql` |
+   | brew       | `brew install postgresql` |
```
NickLaMuro, line 77: I wonder if it makes sense to instead suggest installing `postgresql@10` `brew`, and equivalent for `dnf`/`yum`/`apt`.  Not sure if it is worth doing in this PR, but it is what I do on my end to stay consistent with the version we use.

NickLaMuro (approved):
Looks like fine changes to me.  Left a suggestion, but I think it is pedantic, so probably not worth doing now.

developer_setup.md
kbrock: I too peg the version - I like nick's suggestion

Fryguy: yeah, I was debating that one myself.  I'm not sure how to peg the version in other systems.  @bdunne @agrare ?

Fryguy: I updated this one, but would like to update them all if possible.

NickLaMuro: That is why I was kinda of suggesting solving this another time.  I don't have familiarity with the other package managers like I do with `brew`, so I couldn't provide a solid answer, and I figured we should be consistent with this documentation.

agrare: Debian/Ubuntu it would be `apt install postgresql-10`

It depends on the version of your distro though, for example the latest stable version of debian `postgresql-11` is the oldest version available and I have `postgresql-12` and `postgresql-13` installed.  You would have to enable the prior version's apt source to get `apt install postgresql-10` to work.

Fryguy: Yeah this is why we avoided the version numbers 😆

agrare: Similar for Fedora, `dnf install postgresql-11` and 11 is the oldest version available on Fedora 33

Fryguy: I can just put 11 across the board in the docs.

agrare: > Yeah this is why we avoided the version numbers laughing

Haha yes, for devs it is probably better to have MIQ work with the latest version of PG (e.g. whatever `apt install postgresql` would bring in) even if it isn't "supported"

Fryguy: Thanks updated.  I also added a Min/Max to the initial table of versions, since installing 11 would be weird otherwise.

developer_setup.md
```diff
@@ -2,156 +2,151 @@
...
-  ```bash
-  sudo pg_dropcluster --stop 10 main
-  sudo pg_createcluster -e UTF-8 -l C 10 main -- --auth trust --username root
-  sudo pg_ctlcluster 10 main start
-  ```
+     `brew` will configure the cluster automatically, but you will need to create the user.
 
-* macOS
+     ```bash
+     brew services start postgresql@10
```
NickLaMuro, line 101: Still have `10` as a version number here.

developer_setup.md
```diff
@@ -2,156 +2,151 @@
...
-* Fedora / CentOS
+   * Debian / Ubuntu
 
-  ```bash
-  sudo PGSETUP_INITDB_OPTIONS='--auth trust --username root --encoding UTF-8 --locale C' postgresql-setup --initdb
-  ```
+     ```bash
+     sudo pg_dropcluster --stop 10 main
+     sudo pg_createcluster -e UTF-8 -l C 10 main -- --auth trust --username root
+     sudo pg_ctlcluster 10 main start
```
NickLaMuro, line 93: Same here.

NickLaMuro (changes requested):
Few spots where we are still using PG 10 for version numbers (since it seems like you went with pegging `11` in the docs now).

Fryguy:
Thanks @NickLaMuro Updated.

NickLaMuro (approved):
Good from my perspective.

Wouldn't mind are resident "Debian distro expert", @agrare , confirm that the suggestions for `pg_dropcluster`, `pg_createcluster`, and `pg_ctlcluster` are version numbers for PG, and not some other `10` that was unrelated, but I couldn't see what else that might be off the top of my head.

chessbyte:
Will merge after approval from @agrare and/or @bdunne regarding Linux distros.

agrare:
> confirm that the suggestions for pg_dropcluster, pg_createcluster, and pg_ctlcluster are version numbers for PG, and not some other 10 that was unrelated

:+1: That's correct, `pg_dropcluster [--stop] cluster-version cluster-name` [[ref]](https://www.mankier.com/1/pg_dropcluster)

agrare (approved):

developer_setup.md
bdunne: @carbonin has suggested using https://www.postgresql.org/download/linux/redhat/ in the past

developer_setup.md
```diff
@@ -2,156 +2,151 @@
...
-| dnf        | `sudo dnf -y install postgresql-server` |
-| yum        | `sudo yum -y install postgresql-server` |
-| apt        | `sudo apt -y install postgresql` |
-| brew       | `brew install postgresql` |
-| containers | N/A |
-
-###### NOTE
+   |            |     |
+   | ---------- | --- |
+   | dnf        | `sudo dnf -y install postgresql-11` |
```
bdunne, line 74: This doesn't work for me on Fedora 33

agrare: Ah yeah seems like you can't install a class of versions like you can on deb, probably best to leave this as just `dnf -y install postgresql` then @bdunne ?

bdunne: Yes, `sudo dnf -y install postgresql-server`.  I assume it will be the same for `yum` as well.

Fryguy: Ok I put back the postgresql-server for dnf and yum

Fryguy:
@bdunne this is ready to go

bdunne (approved):

PR #459, opened 2021-04-14
Moved Git Workflow link up for clearer steps

MelsHyrule:
Originally this assumes the user already has an SSH key setup and does not emphasize that we use forks for our development. This is a slight rephrasing and reordering of existing content.


Fryguy (approved):

chessbyte:
@MelsHyrule congratulations on your first ManageIQ PR.  Thank you for your contribution!

PR #463, opened 2021-05-17
Add a simple targeted refresh getting started guide

agrare:
A follow-up to the `Writing a provider from scratch`, a quick guide on how to add targeted refresh to your provider.

### DIFF
--- a/providers/targeted_refresh.md
+++ b/providers/targeted_refresh.md
@@ -1,0 +1,180 @@
+## Targeted Refresh
+
+If you followed the [Creating a new provider](writing_a_new_provider.md) guide then you already have a provider with full refresh working.  Or maybe you're working on an existing provider that only has full refresh and you'd like to add targeted refresh to it.
+
+### What is targeted refresh?
+
+First we should get some technicalities out of the way.  All refreshes have a target, "full" refresh is when the target is the top-level EMS and "targeted" refresh is when the target is anything else (for example a single VM).
+
+The primary benefit of targeted refresh is speed.
+
+Even a moderately large provider will start to take minutes to collect and save all of the inventory.  Doing this simply to pick up a single new VM is extremely wasteful.  Also only one refresh is able to run at a time due to concurrency concerns.  This means if a full refresh takes 10 minutes and you queue up your refresh request right after it starts, then the "latency" to getting your change into the database could take almost 20 minutes.  By contrast fetching a single resource like a VM and related inventory is extremely fast.
+
+There are two main differences between targeted refresh and full refresh:
+
+1. What is collected
+2. How inventory is deleted
+
+Targeted refresh only fetches inventory from the provider for the targets requested.  This cuts down on API calls and uses less memory.
+
+As for deletions, the most significant difference with targeted refresh is the "scope".  A full refresh considers the entire association (e.g.: `ems.vms.all`) to be in scope, so if something is in the association but not in the collected inventory it will be deleted.
+
+Targeted refresh requires scope to be limited, otherwise everything besides what you targeted would be deleted.  Most of this is abstracted away from the provider author but it is important to understand the fundamentals.  This is covered in way more detail in the [Provider Refresh](refresh.md) guide.
+
+### Getting Started
+
+The first thing to do is to set up the new classes that we need.  We'll need to add a `TargetCollection` class under the `Collector` and `Persister`.
+
+```ruby
+class ManageIQ::Providers::AwesomeCloud::Inventory::Collector::TargetCollection < ManageIQ::Providers::AwesomeCloud::Inventory::Collector
+end
+```
+
+```ruby
+class ManageIQ::Providers::AwesomeCloud::Inventory::Parser::TargetCollection < ManageIQ::Providers::AwesomeCloud::Inventory::Parser
+end
+```
+
+```ruby
+class ManageIQ::Providers::AwesomeCloud::Inventory::Persister::TargetCollection < ManageIQ::Providers::AwesomeCloud::Inventory::Persister
+  # Indicates that this is a targeted refresh
+  def targeted?
+    true
+  end
+
+  # This tells the Persister what to do with references, specifically this will
+  # lookup references for inventory objects which aren't in the collection which
+  # is required for targeted refresh.
+  def strategy
+    :local_db_find_missing_references
+  end
+end
+```
+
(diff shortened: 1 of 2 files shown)


providers/targeted_refresh.md
```diff
@@ -0,0 +1,180 @@
...
+  # Indicate that this is a targeted refresh
+  def targeted?
+    true
+  end
+
+  # This tells the Persister what to do with references, specifically this will
+  # lookup references for inventory objects which aren't in the collection which
+  # is required for targeted refresh.
+  def strategy
+    :local_db_find_missing_references
```
agrare, line 49: NOTE: these strategy names are not very intuitive, wonder if we can default the strategy to this if `targeted?` is true so the provider author doesn't need to do this.

providers/targeted_refresh.md
```diff
@@ -0,0 +1,180 @@
...
+    super
+
+    parse_targets!
+  end
+
+  ...
+
+  private
+
+  def parse_targets!
```
agrare, line 103: NOTE: the bulk of these next three methods can probably be moved to the core collector class.  The only thing that we need from the provider is what targets are supported (e.g. MiqTemplate and Vm in this case) but that can be refactored into a method like `supported_targets`

providers/targeted_refresh.md
```diff
@@ -0,0 +1,180 @@
...
+
+If you followed the [Creating a new provider](writing_a_new_provider.md) guide then you already have a provider with full refresh working.  Or maybe you're working on an existing provider that only has full refresh and you'd like to add targeted refresh to it.
+
+### What is targeted refresh?
+
+First we should get some technicalities out of the way.  All refreshes have a target, "full" refresh is when the target is the top-level EMS and "targeted" refresh is when the target is anything else (for example a single VM).
+
+The primary benefit of targeted refresh is speed.
+
+Even a moderately large provider will start to take minutes to collect and save all of the inventory.  Doing this simply to pick up a single new VM is extremely wasteful.  Also since only one refresh is able to run at a time due to concurrency concerns, if a full refresh takes 10 minutes and you queue up your refresh request right after it starts the "latency" to getting your change into the database could take almost 20 minutes.  By contrast fetching a single resource like a VM and related inventory is extremely fast.
```
Fryguy, line 11:
```suggestion
Even a moderately large provider will start to take minutes to collect and save all of the inventory.  Doing this simply to pick up a single new VM is extremely wasteful.  Also since only one refresh is able to run at a time due to concurrency concerns, if a full refresh takes 10 minutes and you queue up your refresh request right after it starts, the "latency" to getting your change into the database could take almost 20 minutes.  By contrast fetching a single resource like a VM and related inventory is extremely fast.
```

Fryguy: For some reason I couldn't read this sentence without that one comma :)

providers/targeted_refresh.md
```diff
@@ -0,0 +1,180 @@
...
+```
+
+```ruby
+class ManageIQ::Providers::AwesomeCloud::Inventory::Parser::TargetCollection < ManageIQ::Providers::AwesomeCloud::Inventory::Parser
+end
+```
+
+```ruby
+class ManageIQ::Providers::AwesomeCloud::Inventory::Persister::TargetCollection < ManageIQ::Providers::AwesomeCloud::Inventory::Persister
+  # Indicate that this is a targeted refresh
```
Fryguy, line 40:
```suggestion
  # Indicates that this is a targeted refresh
```

providers/targeted_refresh.md
```diff
@@ -0,0 +1,180 @@
...
+    :allow_targeted_refresh: true
+```
+
+We should now have all of the scaffolding in place to start implementing the targeted collector.  The goal is that the parser and persister are basically identical in full and targeted refresh and only the collector has to change.
+
+### Targeted Collection
+
+Now that we have the basics down we can begin writing the targeted collector.  Again this is where the majority of the work will be as long as we've done everything correctly.
+
+First we have to override each collection that is defined in the base Collector:
```
Fryguy, line 69: This is an interesting step... if someone adds a new method to the base, they have to remember to "stub" it in the targeted?  That feels error prone, and backwards, somehow.  Not sure how to "fix" that, though.

providers/targeted_refresh.md
agrare: Bit of a run-on already, I'll split it up into two sentences.

providers/targeted_refresh.md
agrare: Yeah I agree, as I was writing this I started looking at moving the "full" out of the base collector class so that targeted isn't inheriting from it and accidentally call the full method.

providers/targeted_refresh.md
```diff
@@ -0,0 +1,180 @@
...
+EmsRefresh.refresh(target)
+ems.reload
+ems.vms.find_by(:ems_ref => "1234")
+```
+
+### Conclusion
+
+And that's all you need to get started!  Targeted refresh is most helpful when paired with an event catcher which we will cover next.
+
+Even without an event catcher you can use this e.g. in a provisioning workflow after the VM is created in-lieu of the full refresh as this can greatly reduce the time each provision request takes.
```
Fryguy, line 180:
```suggestion
Even without an event catcher you can use this (e.g. in a provisioning workflow after the VM is created) in lieu of the full refresh, as this can greatly reduce the time each provision request takes.
```

Fryguy (approved):

Fryguy:
Couple of minor grammatical things, but looks good

providers/targeted_refresh.md
agrare: :+1: done

providers/targeted_refresh.md
agrare: :+1: done

providers/targeted_refresh.md
agrare: :+1: done

providers/targeted_refresh.md
```diff
@@ -0,0 +1,180 @@
...
+EmsRefresh.refresh(target)
+ems.reload
+ems.vms.find_by(:ems_ref => "1234")
+```
+
+### Conclusion
+
+And that's all you need to get started!  Targeted refresh is most helpful when paired with an event catcher which we will cover next.
+
+Even without an event catcher you can use this (e.g. in a provisioning workflow after the VM) is created in lieu of the full refresh, as this can greatly reduce the time each provision request takes.
```
Fryguy, line 180:
```suggestion
Even without an event catcher you can use this (e.g. in a provisioning workflow after the VM is created) in lieu of the full refresh, as this can greatly reduce the time each provision request takes.
```

providers/targeted_refresh.md
chessbyte: All too often, explaining something clarifies things not only for the audience, but for the explainer as well.  The more of these guides we write, the more pluggable our providers will become as we all notice code that seems to be needless or unintuitive.

PR #466, opened 2021-05-19
Add a Provider Events getting started guide

agrare:


providers/events.md
```diff
@@ -0,0 +1,257 @@
...
+(Native Provider)  --->   (Event Catcher Thread)  ---->  (Event Catcher Parser)  ----->  (MiqEventHandler)  -> (VMDB)
+               Emits an event                  Internal Queue                 MiqQueue.put
+```
+
+### Writing an Event Catcher
+
+First we need to set up the worker scaffolding so that we can run our event catcher worker.
+
+`config/settings.yml`
+```
```
Fryguy, line 33:
```suggestion
```yaml
```

providers/events.md
```diff
@@ -0,0 +1,257 @@
...
+    event_hash[:vm_uid_ems] = event.dig(:vm, :id)
+    event_hash[:vm_ems_ref] = event.dig(:vm, :id)
+    event_hash[:vm_name]    = event.dig(:vm, :name)
+  end
+end
+```
+
+And with that you should be able to catch events and have them saved to the database!
+
+To test this out you can run `simulate_queue_worker` in a rails console (this will act as your `MiqEventHandler`) and then you can run you event catcher with `run_single_worker`:
```
Fryguy, line 181:
```suggestion
To test this out you can run `simulate_queue_worker` in a rails console (this will act as your `MiqEventHandler`) and then you can run your event catcher with `run_single_worker`:
```

Fryguy (approved):
Couple of minor things before merge.

providers/events.md
agrare: :+1: done

providers/events.md
agrare: :+1: done

chessbyte:
Does it make sense to link the Events Getting Started Guide and the other recent guides to [this](https://github.com/ManageIQ/guides/blob/master/providers/dev-guide.md)?  Let's make it easier for provider authors to navigate our developer docs.

PR #467, opened 2021-06-18
Add guide on testing logical replication with migrations

jrafanie:
From discussion and testing mentioned in https://github.com/ManageIQ/manageiq/pull/21259

### DIFF
--- a/logical_replication_migrations.md
+++ b/logical_replication_migrations.md
@@ -3,7 +3,7 @@
 ## Introduction
 
 ManageIQ uses logical replication to provide central administrative functions over objects in other
-database regions.  In order to do this, we use postgresql's logical replication to setup
+database regions.  In order to do this, postgresql's logical replication is used to setup
 publications for specific tables in remote regions and subscriptions for each in the central or
 global region.
 
@@ -13,7 +13,7 @@ database it resides.  HA, on the other hand, provides for failover by promoting 
 primary database fails.
 
 This interconnectedness of multiple databases with publications and subscriptions with ManageIQ
-regions means that the upgrade process is a bit different.  Basically, the central region cannot be
+regions means that the upgrade process is a bit different.  Basically, the global region cannot be
 migrated until all databases it's subscribing to have been migrated.  Fortunately, this logic
 should be mostly automatic.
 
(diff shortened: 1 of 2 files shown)


logical_replication_migrations.md
```diff
@@ -0,0 +1,215 @@
...
+## Introduction
+
+ManageIQ uses logical replication to provide central administrative functions over objects in other
+database regions.  In order to do this, we use postgresql's logical replication to setup
+publications for specific tables in remote regions and subscriptions for each in the central or
+global region.
+
+Note, this is completely different from HA and failover.  Logical replication provides product
+features such as being able to start or stop a virtual machine centrally regardless of which
+database it resides.  HA, on the other hand, provides for failover by promoting a backup when a
```
chessbyte, line 12:
```suggestion
database it resides in.  HA, on the other hand, provides for failover by promoting a backup when a
```

Fryguy:
Hi @jrafanie, this is good.

I think moving forward, I'd like to (not necessarily for this PR)

a. Find a way to do this all within a single database (I think you had found a bug)
  a1. If we can script it / have helper methods in console / move into a Rake task, even better.
b. Change the tense from "I did this, then I did that" to "Do this then do that".

jrafanie:
> a. Find a way to do this all within a single database (I think you had found a bug)
>     a1. If we can script it / have helper methods in console / move into a Rake task, even better.

Yeah, I want to come back to that.  It's such a time sink getting it working.

Good point on the bug.  I don't know that it's a problem for us BUT publications are only visible on the databases providing updates on tables whereas subscriptions are visible anywhere in the cluster, so you can't look at subscriptions to know for sure if you're a "subscriber".

jrafanie:
> (I think you had found a bug)

https://github.com/ManageIQ/manageiq/pull/21286

jrafanie:
> b. Change the tense from "I did this, then I did that" to "Do this then do that".

I'll do that now, then, we can merge it. 🤣 

I mean:  "This will be done now.  Then, this PR will be ready."

jrafanie:
Ok, personal pronouns were removed to improve the writing style.   Additionally, a few places were referencing "central" instead of "global" region so all references to "central" were changed to "global" except for the place where they are described as having the same meaning in ManageIQ logical replication terms.

PR #470, opened 2021-09-22
Add 'blog' as a scope label

NickLaMuro:
Avoids needing an extra label in triage meetings like `enhancement` and `other` when this label exists.

Marking this as WIP until I can validate it locally.


Fryguy:
This should have updated 2 sets of links on the bottom.  Please run bin/generate_triage_links.rb which should do the right thing.

NickLaMuro:
@Fryguy ah, that makes sense.  I just manually edited, so good to know.

triage_process.md
```diff
@@ -48,13 +48,13 @@ The ManageIQ core team holds regular meetings to triage issues and pull requests
   * @mention the author after 1 month of inactivity
   * Close the PR after 3 months of inactivity
 
-[scope labels]:                                             https://www.manageiq.org/docs/guides/labels
+[scope labels]:                                             /docs/guides/labels
```
NickLaMuro, line 51: I can remove this change as it is a separate commit, but when working on https://github.com/ManageIQ/manageiq.org/pull/1021 I noticed this when trying to access the link locally, and I was just brought to `manageiq.org` instead of my local copy.

Minor, but seemed related enough to merge it with this change.

PR #474, opened 2021-10-08
Upgrade to yarn 3

Fryguy:
Part of https://github.com/ManageIQ/manageiq-ui-classic/issues/7903


Fryguy:
I was confused and installing yarn is still required.

PR #478, opened 2021-12-16
Create button guide

gregoryb-at-ibm:
Hi

Is this appropriate for the guides repository?

It's a guide on where to add what code to enable creation of an externally managed item in your EMS.

* Need to determine location and name of file.

### DIFF
--- a/providers/create_operation.md
+++ b/providers/create_operation.md
@@ -1,0 +1,142 @@
+# Guide for adding a "new" button 
+You need to already have an application record and a GUI controller that implements a GUI list view (including templates).
+This guide will tell you how to add a button to the list view that creates new instances of the item your application record represents.\
+In this guide we will implement the actual creation of the record in an EMS provider.\
+If creation of your items doesn't require interaction with an EMS API, you can adjust the process described here to implement it in manageiq core instead.
+
+The guide assumes that 
+* Your application-record is in `app/models/your_item.rb`
+* Your application-record belongs to an EMS. Usually its instances appear in MIQ by being refreshed from the EMS.
+* Your UI controller is in `plugins/manageiq-ui-classic/app/controllers/your_item_controller.rb`
+* Your provider is in `plugins/manageiq-providers-your_provider`
+* A menu for your item's list view is in `plugins/manageiq-ui-classic/app/heplers/application_helpers/toolbar/your_items_center.rb`
+* You have an application record in `plugins/manageiq-providers-your_provider/app/models/manageiq/providers/your-provider/your_item.rb`. 
+
+You will work with the following repositories. 
+* manageiq
+* manageiq-ui-classic 
+* manageiq-api
+* manageiq-providers-your_provider (if you implement the creation in an EMS)
+
+## Application Record
+### Base Application Record
+Make sure that the base application-record in `app/models/your_item.rb` has 
+* `self.create_your_item`
+* `self.raw_create_your_item`
+* `self.create_your_item_queue`
+
+If these aren't present, add these methods to the base class by copying them from some other model in `app/models/`.\
+Adjust as needed to fit your item.
+
+### Provider Application Record
+Make sure that your application-record in `plugins/manageiq-providers-your_provider/app/models/manageiq/providers/your-provider/your_item.rb` extend `app/models/your_item.rb` with Rails STI. 
+
+In the application-record in your provider, you need to add two entries.
+1. `supports :create`
+2. `def self.raw_create_your_item(ext_management_system, options = {})` 
+
+The method `self.raw_create_your_item` performs the actual work of creating.\  
+The options hash will contain what you get from your JS form.\
+In this method you need to call your EMS's API to create a real-world instance of your item.\
+You need to check if the request went through OK. If not, you need to throw an exception.
+If the request went OK, you will want to schedule a refresh so that the change will be reflected in MIQ DB.
+
+## Adding support for the new operation 
+In `app/models/ext_management_system.rb` add `supports_attribute :supports_create_your_item, :child_model => "YourItem"`.
+
+## Toolbar Button
+#### Add button to the menu
+Find the menu file in `plugins/manageiq-ui-classic/app/heplers/application_helpers/toolbar/your_items_center.rb`.\
+Add a button into the menu for creating your items. The button needs to be called `:your_item_new`. You can copy a `..._new` button from another menu and adjust to fit your item.
+
+**Note** Usually we only add "new" button into the list view page and not into a detailed view.\
+If you want to also add a "new" button inside you textual summary page you need to do modification to `plugins/manageiq-ui-classic/app/heplers/application_helpers/toolbar/your_item_center.rb`. But this is not part of this guide.
+
+#### Add button class
+Create a button class to handle "new" operation. 
+It needs to be in `manageiq-ui-classic/app/helpers/application_helper/button/your_item_new.rb`. You can copy another button and adjust it.
+
+## JS Form
+You need to create a form-folder in `manageiq-ui-classic/app/javascript/components/your-item-form/`. With `index.jsx` and `your-item-form.schema.js`.
+You can copy from another form and adjust as needed.
+
+Here you determine the GUI form the user will see when creating a new instance of your item. I won't go into the details of it in this guide.\
+When developing, it's safest to start with a minial form just to see that it's working.
+
+Note that the fields on your form will constitute the `options` argument for your `self.raw_create_your_item` method.
+
+After you add your form, you need to register it in `component-definitions-common.js`.
+
+## The UI Controller
+In your controller in `plugins/manageiq-ui-classic/app/controllers/your_item_controller.rb` you need to add the `new` method.\
+You can copy it from another controller and adjust as needed.
+
+Another thing you need in your controller is a method called `specific_buttons(pressed)`. If you don't have it already, copy it from another controller.\
+The method consists of a `case`.\
+Add the following `when` to the case. If you have just copied the method from another controller you can remove all other `when`s.
+```ruby
+when 'your_item_new'
+  javascript_redirect(:action => 'new')
+end
+```
+
+## haml template for 'new' operation
+Create `new.html.haml` under `plugins/manageiq-ui-classic/app/views/your_item/`. You can copy it from another item and adjust as needed.
+Make sure to point at your JS Form which you defined above `react 'YourItemForm'`
+
+## UI routes
+In `manageiq-ui-classic/config/routes.rb`, you need to find the section for your_item. Then you need to make sure there's `new` in the get section, and `button` in the post section.
+
+## Product Feature
+Find your item's product feature list in `manageiq/db/fixtures/miq_product_features.yml`. 
+Make sure that it has a section for 'add', as below
+```yaml
+- :name: Your Item   
+  ...
+  - :name: Modify
+    :children:    
+    - :name: Add
+      :description: Add A Your Item
+      :feature_type: admin
+      :identifier: your_item_new
+```
+
+## API Controller
+If there isn't one, create a controller for your item in `manageiq-api/app/controllers/api/your_items_controller.rb`. You can copy one of the existing controllers and adjust to your item.
+
+Make sure that there is a `create_resource` method in the api controller. If not, copy if from some other controller and adjust to fit your item.\
+This method is supposed to call the method`create_your_item_queue` on the base application-record.\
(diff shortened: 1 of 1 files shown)


Fryguy:
> Is this appropriate for the guides repository?

yes, this is the place!

However, I wonder if this might be better in the provider plugins directory as opposed to UI.  @agrare Thoughts on organization?

agrare:
> However, I wonder if this might be better in the provider plugins directory

Yeah this could actually be a great next "article" in the writing a new provider series, e.g. https://github.com/ManageIQ/guides/blob/master/providers/writing_a_new_provider.md#next-steps

ui/flow_of_allowing_to_add_models.md
```diff
@@ -0,0 +1,123 @@
...
+Make sure the database table for the application-record has a "type" column, so that STI can work properly.\
+If it doesn't have such a column, you'll need to add a migration file to `manageiq-schema` and create the column.    
+It will look something like
+```ruby
+class AddTypeToYourItems < ActiveRecord::Migration[6.0]
+  def change
+    add_column :your_items, :type, :string
+  end
+end
+```
```
agrare, line 40: This is usually handled when adding the new model, not when adding create support.  Maybe better to split up the "How to add a new model" into its own guide which would cover [[lines 22 - 40]](https://github.com/ManageIQ/guides/pull/478/files#diff-7b7fe923f33307d7a91073faa2e44c3b61e1c9c60a8e0db971e67ea7e7a960c6R22-R40)

ui/flow_of_allowing_to_add_models.md
```diff
@@ -0,0 +1,123 @@
...
+
+### Support create
+In the application-record in your provider, you need to add two entries.
+1. `supports :create`
+2. `def self.raw_create_your_item(ext_management_system, options = {})` 
+
+The method `self.raw_create_your_item` performs the actual work of creating.\  
+The options dict will contain what you get from your JS form.\
```
agrare, line 48: `options dict`

In ruby this is more commonly called a hash, python has dictionaries

ui/flow_of_allowing_to_add_models.md
```diff
@@ -0,0 +1,123 @@
...
+In this method you need to call your EMS's API to create a real-world instance of your item.\
+You need to check if the request went through OK. If not, you need to throw an exception.
+If the request went OK, you will want to schedule a refresh so that the change will be reflected in MIQ DB.
+
+## Adding support for the new operation 
+1. In `app/models/ext_management_system.rb` add `supports_attribute :supports_create_your_item, :child_model => "YourItem"`
+2. In `manageiq-providers-autosde/app/models/manageiq/providers/your-provider/your_manager.rb`, add `supports :add_your_item`
```
agrare, line 55: `manageiq-providers-autosde` => best to leave this generic as well like `manageiq-providers-your_provider`

agrare: Actually with ``supports_attribute :supports_create_your_item, :child_model => "YourItem"` the correct way to do this would be:
`manageiq-providers-your_provider/app/models/manageiq/providers/your-provider/your_manager/your_thing.rb` and add `supports :create`

ui/flow_of_allowing_to_add_models.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Toolbar Button
```
agrare, line 57: cc @kavyanekkalapu can you review the UI parts of this?

ui/flow_of_allowing_to_add_models.md
```diff
@@ -0,0 +1,123 @@
...
+
+## API Controller
+If there isn't one, create a controller for your item in `manageiq-api/app/controllers/api/your_items_controller.rb`. You can copy one of the existing controllers and adjust to your item.
+
+Make sure that there is a `create_resource` method in the api controller. If not, copy if from some other controller and adjust to fit your item.\
+This method is supposed to call the method`create_your_item_queue` on the base application-record.
+
+An API controller with a good example of `create_resource` method is `auth_key_pairs_controller.rb`, so you can copy-paste from there and adjust as needed.
+
+If you are adding a new controller, you'll need to add routes to in `manageiq-api/api.yml`. Copy from one of the existing routes and adjust.
```
agrare, line 108: If you're adding create to the collection you'll also need to edit `api.yml`, think it'd be `:verbs: *gpppd` but that always confuses me :laughing: assume it is `GET, PUT, PATCH, POST, DESTROY` but what if you only have `GET, POST` is it `*gp`?  How do we know that's not `GET, PATCH`? cc @Fryguy

agrare: NOTE you also might need to create a new product feature

ui/flow_of_allowing_to_add_models.md
```diff
@@ -0,0 +1,123 @@
...
+
+### API Controller Specs
+The spec file needs to be in `plugins/manageiq-api/spec/requests/your_items_spec.rb`.\
+If a spec file doesn't exist for the api-controller, such as when you have just created a new api-controller, create a new one.  
+
+Add the following tests to the spec file. You can copy-paste them from other specs and adjust to your item.
+1. Good path where a POST succeeds in creating a new item.
+2. A create request without a valid ems-id should fail.
+3. A create request for a user without an appropriate role should fail.
```
agrare, line 117: This would be a good place to mention the product_feature/identifier that you have to authorize the user for in the specs

agrare:
I love this @gregoryb-at-ibm thanks so much for putting the whole flow together!

ui/flow_of_allowing_to_add_models.md
kavyanekkalapu: @gregoryb-at-ibm UI section Looks good to me.

ui/flow_of_allowing_to_add_models.md
Fryguy: yes, `gpppd` is `GET POST PUT PATCH DESTROY`.  The star at the beginning of `*gpppd` is a YAML anchor that points to the definition.  See https://github.com/ManageIQ/manageiq-api/blob/master/config/api.yml#L6-L31  We know it's not `GET PATCH`, because `GET PATCH` is not defined 😉

agrare: :+1: so if you wanted to implement only GET and PATCH you could call them out and just not use the shorthand

Fryguy: If it helps (or for others that aren't aware of YAML anchors)...

```ruby
[1] pry(main)> require "yaml"
=> true
[2] pry(main)> x = {:target => {:a => 1, :b => 2}}
=> {:target=>{:a=>1, :b=>2}}
[3] pry(main)> x[:reference] = x[:target]
=> {:a=>1, :b=>2}
[4] pry(main)> x[:reference].object_id == x[:target].object_id
=> true
[5] pry(main)> puts x.to_yaml
---
:target: &1
  :a: 1
  :b: 2
:reference: *1
=> nil
[6] pry(main)> y = YAML.load(x.to_yaml)
=> {:target=>{:a=>1, :b=>2}, :reference=>{:a=>1, :b=>2}}
[7] pry(main)> y[:reference].object_id == y[:target].object_id
=> true
```

chessbyte: so we created this nomenclature to minimize typing??

chessbyte: anchors are also used by Rails in `config/database.yml`, if I recall correctly

Fryguy: It minimizes _a lot_ of duplication, but yes, it was created to declutter the file.  It also ensures that only known "sets" are chosen.

Fryguy: For reference:

```ruby
[8] pry(main)> YAML.load_file("config/api.yml").deep_dup.to_yaml.lines.count
=> 5457
[9] pry(main)> File.read("config/api.yml").lines.count
=> 5068
```

The first line is fully expanded, the second line is the current, so we're talking 389 lines or 7.5% bigger

gregoryb-at-ibm: @agrare, @Fryguy 
I added instructions about adding `create` to YAML. Could you see that it's correct and complete?

Also, could you help me with instructions for adding an appropriate product feature?

ui/flow_of_allowing_to_add_models.md
gregoryb-at-ibm: I've added a section about product features. Could you check it?

ui/flow_of_allowing_to_add_models.md
```diff
@@ -0,0 +1,144 @@
...
+* manageiq-providers-your_provider (if you implement the creation in an EMS)
+
+## Application Record
+### Base Application Record
+Make sure that the base application-record in `app/models/your_item.rb` has 
+* `self.create_your_item`
+* `self.raw_create_your_item`
+* `self.create_your_item_queue`
+
+If these aren't present, add these methods to the base record by copying them from some other model in `app/models/`.
```
agrare, line 29: suggest changing `record` -> `base class` or `base model`

ui/flow_of_allowing_to_add_models.md
```diff
@@ -0,0 +1,144 @@
...
+
+### Provider Application Record
+Make sure you have an application-record in `plugins/manageiq-providers-your_provider/app/models/manageiq/providers/your-provider/your_item.rb`.
+It should extend `app/models/your_item.rb` with Rails STI.
```
agrare, line 33: Do you want to remove [[this]](https://github.com/ManageIQ/guides/pull/478/files#diff-7b7fe923f33307d7a91073faa2e44c3b61e1c9c60a8e0db971e67ea7e7a960c6R13) from the assumptions if you're also covering it here?  Think it should be in one place or the other but not both

ui/flow_of_allowing_to_add_models.md
```diff
@@ -0,0 +1,144 @@
...
+
+The method `self.raw_create_your_item` performs the actual work of creating.\  
+The options hash will contain what you get from your JS form.\
+In this method you need to call your EMS's API to create a real-world instance of your item.\
+You need to check if the request went through OK. If not, you need to throw an exception.
+If the request went OK, you will want to schedule a refresh so that the change will be reflected in MIQ DB.
+
+## Adding support for the new operation 
+1. In `app/models/ext_management_system.rb` add `supports_attribute :supports_create_your_item, :child_model => "YourItem"`
+2. In `manageiq-providers-your_provider/app/models/manageiq/providers/your-provider/your_item.rb`, add `supports :create`
```
agrare, line 47: Only need to talk about adding `supports :create` once, [[ref]](https://github.com/ManageIQ/guides/pull/478/files#diff-7b7fe923f33307d7a91073faa2e44c3b61e1c9c60a8e0db971e67ea7e7a960c6R36)

ui/flow_of_allowing_to_add_models.md
```diff
@@ -0,0 +1,144 @@
...
+Create `new.html.haml` under `plugins/manageiq-ui-classic/app/views/your_item/`. You can copy it from another item and adjust as needed.
+Make sure to point at your JS Form which you defined above `react 'YourItemForm'`
+
+## UI routes
+In `manageiq-ui-classic/config/routes.rb`, you need to find the section for your_item. Then you need to make sure there's `new` in the get section, and `button` in the post section.
+
+## Product Feature
+Find your item's product feature list in `manageiq/db/fixtures/miq_product_features.yml`. 
+Make sure that it has a section for 'add', as below
+```shell
```
agrare, line 95: Really minor, but this should be `yaml` not `shell`

agrare:
Looking good @gregoryb-at-ibm just a few minor comments then I'll merge!

ui/flow_of_allowing_to_add_models.md
gregoryb-at-ibm: > Looking good @gregoryb-at-ibm just a few minor comments then I'll merge!

I think I fixed all the comments?

Before you merge we should decide on the file's name, and where it should go.
And also I'd like to squash all my commits.

gregoryb-at-ibm:
@agrare 
I think I fixed all the comments?

Before you merge we should decide on the file's name, and where it should go.
Then I'll squash all my commits.

agrare:
> Before you merge we should decide on the file's name, and where it should go.

How about `providers/create_operation.md`

gregoryb-at-ibm:
> > Before you merge we should decide on the file's name, and where it should go.
> 
> How about `providers/create_operation.md`

OK, moved and squashed.

agrare (approved):

PR #491, opened 2022-03-04
syncing appliance auth-openidc config with guide

kbrock:
https://github.com/ManageIQ/manageiq-appliance/blob/master/TEMPLATE/etc/httpd/conf.d/manageiq-external-auth-openidc.conf.erb

- added same site
- changed indentation
- highlighted changes from our production setup

Thoughts:
- assuming the crypto passphrase is probably fine being different. Would it be that much work to make the same?
- do we want to add tp production: `OIDCOAuthIntrospectionEndpointAuth` (have not researched yet)
- ssl changes are good / not for production
- do we want to add to production: `let_admin_in`
- do we want to add to production: X-REMOTE-HEADERS (tricky stuff)

Do you want me to sync with ipa-configuration and look into pods to make sure those are the same too?


external_auth/oidc-httpd-configs/authentication.conf
```diff
@@ -2,51 +2,67 @@ LoadModule auth_openidc_module modules/mod_auth_openidc.so
...
   Allow from env=let_csrf_token_in
   Satisfy Any
-
-  SetEnvIf Authorization '^Basic +YWRtaW46' let_admin_in
+  # different from production
   Allow from env=let_admin_in
+  # /differen
 </LocationMatch>
 
+# different from production
```
Fryguy, line 65: This part shouldn't be different from production

external_auth/oidc-httpd-configs/authentication.conf
```diff
@@ -2,51 +2,67 @@ LoadModule auth_openidc_module modules/mod_auth_openidc.so
...
   Order          Allow,Deny
   Allow from env=let_api_token_in
   Allow from env=let_sys_token_in
   Allow from env=let_csrf_token_in
   Satisfy Any
-
-  SetEnvIf Authorization '^Basic +YWRtaW46' let_admin_in
+  # different from production
   Allow from env=let_admin_in
+  # /differen
```
Fryguy, line 62:
```suggestion
  # /different
```

external_auth/oidc-httpd-configs/authentication.conf
```diff
@@ -2,51 +2,67 @@ LoadModule auth_openidc_module modules/mod_auth_openidc.so
...
-OIDCOAuthRemoteUserClaim     username
-OIDCCacheShmEntrySizeMax     65536
 
+OIDCProviderMetadataURL            <%= oidc_provider_metadata_url %>
+OIDCProviderMetadataURL            https://${HTTPD_AUTH_HOST}:8443/auth/realms/ManageIQ/.well-known/openid-configuration
+OIDCClientID                       ${HTTPD_AUTH_OIDC_CLIENT_ID}
+OIDCClientSecret                   ${HTTPD_AUTH_OIDC_CLIENT_SECRET}
+OIDCRedirectURI                    http://${HTTPD_AUTH_HOST}/oidc_login/redirect_uri
+# different from production
+OIDCCryptoPassphrase               sp-secret
```
Fryguy, line 12: Since this can basically be any string we want...let's just make it match production

external_auth/oidc-httpd-configs/authentication.conf
```diff
@@ -2,51 +2,67 @@ LoadModule auth_openidc_module modules/mod_auth_openidc.so
...
-OIDCProviderMetadataURL      https://${HTTPD_AUTH_HOST}:8443/auth/realms/ManageIQ/.well-known/openid-configuration
-OIDCClientID                 ${HTTPD_AUTH_OIDC_CLIENT_ID}
-OIDCClientSecret             ${HTTPD_AUTH_OIDC_CLIENT_SECRET}
-OIDCRedirectURI              http://${HTTPD_AUTH_HOST}/oidc_login/redirect_uri
-OIDCCryptoPassphrase         sp-secret
-OIDCOAuthRemoteUserClaim     username
-OIDCCacheShmEntrySizeMax     65536
 
+OIDCProviderMetadataURL            <%= oidc_provider_metadata_url %>
+OIDCProviderMetadataURL            https://${HTTPD_AUTH_HOST}:8443/auth/realms/ManageIQ/.well-known/openid-configuration
```
Fryguy, line 7: ?? You can't have this twice.

Fryguy:
>   - assuming the crypto passphrase is probably fine being different. Would it be that much work to make the same?

We can just make them match

>   - do we want to add tp production: OIDCOAuthIntrospectionEndpointAuth (have not researched yet)

If production doesn't have it, then it can probably be removed from here.  Everything should be discoverable via the `OIDCProviderMetadataURL`

>   - do we want to add to production: let_admin_in

let_admin_in is technically gated by the value of  EnableApplicationLocalLogin.  If it's false, then let_admin_in should not be present (which is the default when using OIDC).

>   - do we want to add to production: X-REMOTE-HEADERS (tricky stuff)

Those should already be there.

> Do you want me to sync with ipa-configuration and look into pods to make sure those are the same too?

You mean the ipa configuration guides?  I'm not even sure what's in those and they likely need an overhaul, so I'd say that's a separate effort.

external_auth/oidc-httpd-configs/authentication.conf
kbrock: ok, I'll reencrypt the private key

external_auth/oidc-httpd-configs/authentication.conf
kbrock: and the first one is erb and wouldn't even work here

kbrock:
WIP: these guides are more similar to the manageiq-pods ipa-configuration so I will sync to those values.

external_auth/oidc-httpd-configs/authentication.conf
kbrock: looks like it is the same as pods. leaving it as is

kbrock:
update: ok, synced with pods. this was much more similar

PR #493, opened 2022-04-04
Add a guide for writing provider VCR specs

agrare:
Follow-ups:
* Add spec test examples for existing targeted refresh guide
* Examples for auto-generating inventory with a rake task

### DIFF
--- a/providers/writing_vcr_specs.md
+++ b/providers/writing_vcr_specs.md
@@ -1,0 +1,128 @@
+## Writing VCR Provider Spec Tests
+
+So you've just done a refresh with your new provider, congratulations.  Now it is time to make sure it continues to work by writing some tests.
+
+### VCR
+
+How do you test code which makes API calls without having access to a live system?  There is a ruby gem called [VCR](https://github.com/vcr/vcr) which "records" HTTP API calls and then plays them back later.  This allows us to test code which would normally have to point to a live API without having to stub every single method call.
+
+The results of the recording are stored in the VCR `config.cassette_library_dir` as YAML files with request/response pairs.  VCR will automatically stub the HTTP layer for you, and when it recognizes a request it will replay the saved response for you.  This allows for mostly normal operation while running in an isolated CI system.
+
+### Configuration
+
+If you generated your provider plugin with the `--vcr` flag then the VCR configuration should have already been done for you.  If not simply add the following to the bottom of your `spec/spec_helper.rb` file:
+```ruby
+VCR.configure do |config|
+  config.ignore_hosts 'codeclimate.com' if ENV['CI']
+  config.cassette_library_dir = ManageIQ::Providers::AwesomeCloud::Engine.root.join('spec/vcr_cassettes')
+end
+```
+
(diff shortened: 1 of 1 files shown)


providers/writing_vcr_specs.md
```diff
@@ -0,0 +1,136 @@
+## Writing VCR Provider Spec Tests
+
+So you've just done a refresh with your new provider, congratulations.  Now it is time to make sure it continues to work by writing some tests.
+
+### VCR
+
+How do you test code which makes API calls without having access to a live system?  There is a ruby gem called VCR which "records" HTTP API calls and then plays them back later.  This allows us to test code which would normally have to point to a live API without having to stub every single method call.
```
Fryguy, line 7:
```suggestion
How do you test code which makes API calls without having access to a live system?  There is a ruby gem called [VCR](https://github.com/vcr/vcr) which "records" HTTP API calls and then plays them back later.  This allows us to test code which would normally have to point to a live API without having to stub every single method call.
```

providers/writing_vcr_specs.md
```diff
@@ -0,0 +1,136 @@
...
+
+The results of the recording are stored in the VCR `config.cassette_library_dir` as YAML files with request/response pairs.  VCR will automatically stub the HTTP layer for you, and when it recognizes a request it will replay the saved response for you.  This allows for mostly normal operation while running in an isolated CI system.
+
+### Configuration
+
+If you generated your provider plugin with the `--vcr` flag then the VCR configuration should have already been done for you.  If not simply add the following to the bottom of your `spec/spec_helper.rb` file:
+```ruby
+VCR.configure do |config|
+  config.ignore_hosts 'codeclimate.com' if ENV['CI']
+  config.cassette_library_dir = File.join(ManageIQ::Providers::AwesomeCloud::Engine.root, 'spec/vcr_cassettes')
```
Fryguy, line 17: Not for this PR, but I was surprised this wasn't:

```suggestion
  config.cassette_library_dir = ManageIQ::Providers::AwesomeCloud::Engine.root.join('spec/vcr_cassettes')
```

providers/writing_vcr_specs.md
```diff
@@ -0,0 +1,136 @@
...
+Then create a `config/secrets.yml` file (this file will not be committed and should be in your .gitignore):
+```yaml
+---
+test:
+  awesome_cloud:
+    access_key: "YOUR_REAL_ACCESS_KEY"
+    secret_key: "YOUR_REAL_SECRET_KEY"
+```
+
+Then add the following to your `VCR.configure` block in `spec/spec_helper.rb`:
```
Fryguy, line 45: You might want to minimize this to only the "new" parts (thought I understand why you might want it all for context).

Fryguy (approved):

providers/writing_vcr_specs.md
agrare: Yeah I thought about that but didn't want people just putting it at the end of the file outside of the block but guess there is just as good a chance they'll duplicate the whole thing.  I'll add a comment about what line specifically to add it after

PR #501, opened 2022-07-29
Update developer_setup.md

iv1111:
- note about MacOS on silicon chip and Homebrew
- see [Homebrew/brew#13481](https://github.com/Homebrew/brew/issues/13481)

### DIFF
--- a/developer_setup.md
+++ b/developer_setup.md
@@ -41,6 +41,9 @@
    | apt  | `sudo apt -y install build-essential libffi-dev libpq-dev libxml2-dev libcurl4-openssl-dev cmake python libssh2-1-dev` |
    | brew | `brew install cmake libssh2` |
 
+   **Note**: Users with MacOS running on Apple M1 CPU might need to specify location of the `Homebrew` libraries explicitly. 
+   If build steps below fail for you then run `export LIBRARY_PATH="$LIBRARY_PATH:$(brew --prefix)/lib"`and retry the failing build step.
+
 ### Services
 
 ManageIQ requires a memcached instance for session caching and a PostgreSQL database for persistent data storage.
--- a/developer_setup.md
+++ b/developer_setup.md
@@ -43,6 +43,7 @@
 
    **Note**: Users with MacOS running on Apple M1 CPU might need to specify location of the `Homebrew` libraries explicitly. 
    If build steps below fail for you then run `export LIBRARY_PATH="$LIBRARY_PATH:$(brew --prefix)/lib"`and retry the failing build step.
+   See here for additional information: https://github.com/Homebrew/brew/issues/13481
 
 ### Services
 


iv1111:
FYI @agrare @jrafanie @Fryguy.

developer_setup.md
```diff
@@ -41,6 +41,10 @@
    | apt  | `sudo apt -y install build-essential libffi-dev libpq-dev libxml2-dev libcurl4-openssl-dev cmake python libssh2-1-dev` |
    | brew | `brew install cmake libssh2` |
 
+   **Note**: Users with MacOS running on Apple M1 CPU might need to specify location of the `Homebrew` libraries explicitly. 
+   If build steps below fail for you then run `export LIBRARY_PATH="$LIBRARY_PATH:$(brew --prefix)/lib"`and retry the failing build step.
```
agrare, line 45: > and retry the failing build step.

Wonder if we should move this up before the `brew install` steps so that people can avoid "try it, if it fails do this, then try it again"

agrare: > `export LIBRARY_PATH="$LIBRARY_PATH:$(brew --prefix)/lib"`

If this is similar to linux then this is only a temporary solution, would users also have to add this to a mac equivalent of a .bashrc file?  Otherwise would this just be lost when the close the terminal/logout/reboot?

jrafanie: > Wonder if we should move this up before the `brew install` steps so that people can avoid "try it, if it fails do this, then try it again"

I'm fine with having this as a note afterwards as it is an exceptional case for now.  Maybe including the "ld: library not found" error message would better" would help direct people to only apply this change if you're having such a problem.  Additionally, exact error messages are what people search for so we want them to find this.

If apple CPUs become the most common mac platform for us and this is still not resolved by homebrew by then, then we can proactively make it a step to follow by default on macs.

agrare: I wonder if we could check for this in bin/install ? If the platform is Darwin check if brew is in the library path.

I imagine M1 CPUs will become the predominant Mac architecture in a few years

MelsHyrule: Note: installing the pg gem didnt work without including this `export PATH=$PATH:/opt/homebrew/Cellar/postgresql@11/11.16/bin` in zsh

agrare: @iv1111 did you have to update your `PATH` as well to install the `pg` gem or did you not get that far?

Fryguy: The more succinct version of that would be `export PATH=$PATH:$(brew --prefix postgresql@11)/bin`

iv1111: > @iv1111 did you have to update your `PATH` as well to install the `pg` gem or did you not get that far?

I didn't have that problem.

agrare: Interesting, can you run `which pg_config` ?

iv1111: @agrare The result is: `/opt/homebrew/opt/postgresql@11/bin/pg_config`

iv1111:
@Fryguy @agrare the issue is still relevant. I have accidentally closed this PR, reopening.

Fryguy:
Merging since it covers the issue at hand - we can cleanup or add more details in follow up PRs

PR #506, opened 2022-09-22
Add steps to do refresh on provider in main developer_setup.md 

mahantidinesh7:


developer_setup.md
```diff
@@ -191,6 +191,14 @@ The web UI should now be available at `http://localhost:3000`. The default usern
...
 Note that in this runtime mode we are only verifying that the components are installed correctly.  However, individual workers are not running, so actions that require asynchronous handling will appear to hang forever.  To run the application complete with all workers, you can start with `bundle exec rails evm:start` and stop with `bundle exec rails evm:stop`.  Otherwise, read ahead to learn about running individual workers during development.
 
+If you need to do also the refresh on providers, We recommend running `rails console` (or `rails c`) in a new terminal window and typing in `simulate_queue_worker`. This command in the rails console will do the work if you click in UI on the refresh button. If you want to run the refresh directly from the rails console, you may want to do that with
+
+```ruby
+reload!; EmsRefresh.refresh(ExtManagementSystem.last)
+```
+
+The `reload!` in the command is not necessary if you didn't modify the code.
+
```
agrare, line 201: I'm thinking rather than explain that async work will hang forever we just directly instruct people to either run simulate_queue_worker.

Something like
```suggestion
ManageIQ runs a lot of work asynchronously via background queue workers, to simulate this we recommend running:
```bash
bundle exec rails console
simulate_queue_worker
```
```

agrare:
I've pushed my suggestion to @mahantidinesh7 's fork, since this now has my commit in it I'll assign to someone else

PR #509, opened 2023-01-10
Guide for Adding a New Model

OrGur1987:
During some enhancements we tried to add to the autosde provider, we encountered difficulties in finding the correct way to 
add a model collected from the provider-client to miq inventory (the collect-parse-build-persist flow).

After figuring it out, we've written a short guide for this, and @agrare suggested that we add it to manageiq/guides.
We've found this article about the inventory, which seems like a good place to add examples that might make it easier for future developers. We use `storage_manager` as the provider class in these examples.

### DIFF
--- a/providers/adding_a_new_model.md
+++ b/providers/adding_a_new_model.md
@@ -1,0 +1,106 @@
+# Adding A New Model
+
+This article provides examples for adding a new model to ManageIQ, to represent and manipulate a model from a provider's backend. The article uses the autosde provider plug-in and its `storage_manager` as example. Note that the exact names, details and file paths may differ between providers.  
+
+Related articles which may expand on subjects mentioned here:
+- [dev-guide](./dev-guide.md)
+- [Inventory Collections](./persister/inventory_collections.md)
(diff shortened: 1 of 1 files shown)


agrare:
This is great @OrGur1987 I think this would be great as a new standalone guide in the "writing a provider" series.

Can you pull this out into a `providers/adding_a_new_model.md` and we can cover everything from generating the schema, adding the core model class, and this section that you have here for adding the new model to inventory collection.

I can write the first few steps if you want to collaborate on this branch together

providers/refresh.md
```diff
@@ -138,3 +138,57 @@ persister.vms.build(
   :host => persister.hosts.lazy_find({:uuid => vm.host_uuid}, {:ref => :by_uuid})
 )
 ```
+<br>
```
Fryguy, line 141: This shouldn't be needed in markdown

Fryguy:
Awesome...thanks @OrGur1987 !

agrare (approved):
:+1: this is awesome, will follow-up with some minor modifications but I'll merge as is

PR #514, opened 2023-02-27
Cleanup old/stale postgres version and installation instructions

jrafanie:
Note, the second commit deletes a completely unused doc that's not even linked to from the main dev setup doc.


developer_setup.md
```diff
@@ -76,8 +76,8 @@ ManageIQ requires a memcached instance for session caching and a PostgreSQL data
    | ---------- | --- |
    | dnf        | `sudo dnf -y install postgresql-server` |
    | yum        | `sudo yum -y install postgresql-server` |
-   | apt        | `sudo apt -y install postgresql-11` |
-   | brew       | `brew install postgresql@11` |
+   | apt        | `sudo apt -y install postgresql-13` |
```
jrafanie, line 79: @agrare can you eyeball the debian instructions.  I just replaced numbers.  I hope they follow conventions.

kbrock: I think there is an odd vs even thing here. Don't we want evens?

agrare: @jrafanie yes that is the convention, but there's no guarantee that v13 will be available on the distribution though.  For example no ubuntu LTS version has the postgresql-13 package, 20.04 has postgresql-12 and 22.04 has postgresql-14.  21.10 does have it but that is out of support already.  Most of the time I think developers just figure out what version they have and use that.  Using 13 here might be fine for an example of what to do, but for a truly copy&paste we'd need to do more

kbrock: never mind ==> https://www.postgresql.org/message-id/200006061607.MAA27510%40candle.pha.pa.us

agrare: There is a "virtual package" called `postgresql` that points to the specific version which is what would typically be installed by devs.
If we want 13 specifically we'll need to add docs on how to add e.g. the pgdg repository, if we want to work with the version of postgres that the distro has we'll need to modify e.g. the `sudo pg_dropcluster --stop 13 main` below to determine the installed version

jrafanie: what do you think makes sense for the most obvious future-proof solution?  It seems weird to continually have to revise these instructions

agrare: As long as we're okay with the version not being the same that a MIQ appliance might use (these are developer instructions so seems not as important) then the way to go is 100% to use the virtual package.

Something that uses `pg_lsclusters --json | jq -r '.[].version'` to get the version should work.  If you want to change this to just be
```suggestion
   | apt        | `sudo apt -y install postgresql` |
```

I'll comment on the other lines with commands that work dynamically with the latest installed version

kbrock: Homebrew on the mac has the same thing. We used to point to `postgres`

But since on Homebrew, upgrading the version of ssl can trigger an upgrade of postgres to a different major version, it got too fragile. So we nailed down the version.

I would imagine no other packaging system would do this, so just pointing to `postgres` on debian makes sense to me.

developer_setup.md
```diff
@@ -91,17 +91,17 @@ ManageIQ requires a memcached instance for session caching and a PostgreSQL data
    * Debian / Ubuntu
 
      ```bash
-     sudo pg_dropcluster --stop 11 main
-     sudo pg_createcluster -e UTF-8 -l C 11 main -- --auth trust --username root
-     sudo pg_ctlcluster 11 main start
+     sudo pg_dropcluster --stop 13 main
+     sudo pg_createcluster -e UTF-8 -l C 13 main -- --auth trust --username root
+     sudo pg_ctlcluster 13 main start
```
agrare, line 96:
```suggestion
     pg_version=$(pg_lsclusters --json | jq -r '.[].version' | sort -n | tail -1) # TODO test with multiple installed postgres versions
     sudo pg_dropcluster --stop $pg_version main
     sudo pg_createcluster -e UTF-8 -l C $pg_version main -- --auth trust --username root
     sudo pg_ctlcluster $pg_version main start
```

developer_setup.md
```diff
@@ -91,17 +91,18 @@ ManageIQ requires a memcached instance for session caching and a PostgreSQL data
    * Debian / Ubuntu
 
      ```bash
-     sudo pg_dropcluster --stop 11 main
-     sudo pg_createcluster -e UTF-8 -l C 11 main -- --auth trust --username root
-     sudo pg_ctlcluster 11 main start
+     pg_version=$(pg_lsclusters --no-header | awk '{print $1}' | sort -n | tail -1)
+     sudo pg_dropcluster --stop $pg_version main
+     sudo pg_createcluster -e UTF-8 -l C $pg_version main -- --auth trust --username root
+     sudo pg_ctlcluster $pg_version main start
```
jrafanie, line 97: Ok, I added @agrare's suggestion for debian.  Thanks for the suggestions!

developer_setup.md
agrare: Just for you I'll whack my database to test this ;)
```
$ pg_version=$(pg_lsclusters --no-header | awk '{print $1}' | sort -n | tail -1)
     sudo pg_dropcluster --stop $pg_version main
     sudo pg_createcluster -e UTF-8 -l C $pg_version main -- --auth trust --username root
     sudo pg_ctlcluster $pg_version main start
Creating new PostgreSQL cluster 15/main ...
/usr/lib/postgresql/15/bin/initdb -D /var/lib/postgresql/15/main --no-instructions --encoding UTF-8 --auth trust --username root
The files belonging to this database system will be owned by user "postgres".
This user must also own the server process.

The database cluster will be initialized with locale "en_US.UTF-8".
The default text search configuration will be set to "english".

Data page checksums are disabled.

fixing permissions on existing directory /var/lib/postgresql/15/main ... ok
creating subdirectories ... ok
selecting dynamic shared memory implementation ... posix
selecting default max_connections ... 100
selecting default shared_buffers ... 128MB
selecting default time zone ... America/New_York
creating configuration files ... ok
running bootstrap script ... ok
performing post-bootstrap initialization ... ok
syncing data to disk ... ok
Ver Cluster Port Status Owner    Data directory              Log file
15  main    5432 down   postgres /var/lib/postgresql/15/main C
adam@desktop:~$ pg_lsclusters 
Ver Cluster Port Status Owner    Data directory              Log file
15  main    5432 online postgres /var/lib/postgresql/15/main C
```

Looks good

developer_setup.md
agrare, line 79: Do we want to leave this as `sudo apt -y install postgresql` ?

jrafanie: yeah, I'd rather we specify the versions once at top and give the general instructions down below.  These specific things always get stale so quickly.

agrare: Yeah I thought we wanted to use the virtual package `postgresql` rather than specific versions

agrare (approved):

developer_setup.md
```diff
@@ -76,8 +76,8 @@ ManageIQ requires a memcached instance for session caching and a PostgreSQL data
    | ---------- | --- |
    | dnf        | `sudo dnf -y install postgresql-server` |
    | yum        | `sudo yum -y install postgresql-server` |
-   | apt        | `sudo apt -y install postgresql-11` |
-   | brew       | `brew install postgresql@11` |
+   | apt        | `sudo apt -y install postgresql` |
```
jrafanie, line 79: fixed...now, only mac is different :-)

Fryguy (approved):

PR #518, opened 2023-03-11
add note for global yarn install

dennym:
The global yarn install needs to be set to version 3 instead of the default install of v1.

When `bin/setup` the CiscoIntersight Provider is going to complain and without internal knowledge of yarn its going to be tricky to debug.

more info: https://yarnpkg.com/getting-started/migration#why-should-you-migrate

### DIFF
--- a/developer_setup.md
+++ b/developer_setup.md
@@ -143,6 +143,7 @@ Then install `yarn` - you can find the recommended way for your platform at http
 
 ```bash
 npm install --global yarn
+yarn set version berry
 ```
 
 ### Ruby and Bundler


Fryguy (approved):

Fryguy:
Closing this because this is not accurate.  yarn v1 is generally what's installed globally, and will respect the yarn version defined in the .yarnrc.yml file.  Newer yarn is expected to be vendored in the .yarn directory within the project.  cisco intersight plugin has it vendored correctly here: https://github.com/ManageIQ/manageiq-providers-cisco_intersight/blob/3c74cf9a0adeb6c53c7075598624e1f8a94e9181/.yarn/releases/yarn-3.5.0.cjs

kbrock:
I wonder if there is a way to document the issue though. It takes a bit of googling to understand why running yarn from different directories behaves differently

PR #519, opened 2023-04-10
Instructions on installing kafka

kbrock:
Here is a first stab of kafka installation instructions for developers.

---

Kafka currently supports java versions 8, 11, 17, but has deprecated jdk 8. I include version 19 since that is the version that the mac installs by default.

- Only really tested on the mac.
- Not totally sure if the system wide java command is necessary on the mac.

/cc @nasark @agrare


agrare:
@jrafanie can you confirm this works on MacOS

jrafanie (approved):
LGTM

PR #532, opened 2023-11-27
Add notes running vmware simulator

kbrock:
Here are notes on running the vmware simulator

not sure if we want this in the providers or developer_setup directory

### DIFF
--- a/providers/vmware_simulator.md
+++ b/providers/vmware_simulator.md
@@ -1,0 +1,98 @@
+# installing vmware simulator
+
+```bash
+cd ~/go/
+git clone https://github.com/kbrock/govmomi
+cd govmomi
+git remote add upstream  https://github.com/vmware/govmomi
+git checkout miq_counters_kb
+# client
+pushd govc ; make clean && make && make install ; popd
+# server
+pushd vcsim ; make clean && make && make install ; popd
+ln -s ~/go/bin/vcsim ~/bin # or /usr/local/bin
+ln -s ~/go/bin/govc ~/bin  # or /usr/local/bin
+```
+
+# creating simulator provider
+
+```bash
+echo "127.0.0.1 vc91" > /etc/hosts
+
+# run temporary simulator 
+vcsim -username root -password vmware -app 2 -cluster 2 -dc 1 -ds 2 -folder 2 \
+      -host 2 -nsx 2 -pg 2 -pg-nsx 2 -pod 2 -pool 1 -standalone-host 2 -vm 2 \
+      -l vc91:9191 &
+
+export GOVC_URL=https://root:vmware@vc91:9191/sdk GOVC_INSECURE=true
+govc object.save -d my-vcenter
+# stop stop temporary simulator
+kill %1 # or fg ; control-c
+```
+
+# adding historicals
+
+```xml
+<!--
+  edit my-vcenter/0009-PerformanceManager-PerfMgr.xml
+  find nistoricalInterval
+  -->
+  <propSet>
+    <name>historicalInterval</name>
+    <val xmlns:XMLSchema-instance="http://www.w3.org/2001/XMLSchema-instance" XMLSchema-instance:type="ArrayOfPerfInterval">
+      <PerfInterval>
+        <key>1</key>,
+        <samplingPeriod>300</samplingPeriod>
+        <length>86400</length>
+        <name>Past Day</name>
+        <level>1</level>
+        <enabled>true</enabled>
+      </PerfInterval>
+      <PerfInterval>
+        <key>2</key>
+        <samplingPeriod>1800</samplingPeriod>
+        <length>604800</length>
+        <name>Past Week</name>
+        <level>1</level>
+        <enabled>true</enabled>
+      </PerfInterval>
+      <PerfInterval>
+        <key>3</key>
+        <samplingPeriod>7200</samplingPeriod>
+        <length>2592000</length>
+        <name>Past Month</name>
+        <level>1</level>
+        <enabled>true</enabled>
+      </PerfInterval>
+      <PerfInterval>
+        <key>4</key>
+        <samplingPeriod>86400</samplingPeriod>
+        <length>31536000</length>
+        <name>Past Year</name>
+        <level>1</level>
+        <enabled>true</enabled>
+      </PerfInterval>
+    </val>
+  </propSet>
+```
+# better editing of historicals
+
+TODO: please figure out a way to add historicals without editing the files.
+That way we can just use the vmware simulator container
+
+```bash
+  govc object.collect -s -dump PerformanceManager:PerfMgr historicalInterval
+  govc metric.interval.info
+
+  govc metric.interval.change
+
+  # this one has a default (not sure if related or not)
+  govc object.collect -s -dump PerformanceManager:PerfMgr perfCounter
+```
+
+# run
+
+```bash
+# I dumped to a local directory to add historical capture support
+vcsim -username root -password vmware -load my-vcenter -l vc91:9191
+```


kbrock:
Had tried to default the historical intervals via https://github.com/vmware/govmomi/pull/3191 but wasn't able to get that working.

agrare:
@kbrock running with metrics collection is probably more "niche" compared with what the average developer would use this for?
What if the primary documentation is to `go install github.com/vmware/govmomi/govc@latest` and then `vcsim -username root -password vmware ...` ?  Maybe we put the instructions for historical intervals in the manageiq-providers-vmware README?

providers/vmware_simulator.md
```diff
@@ -0,0 +1,98 @@
...
+# server
+pushd vcsim ; make clean && make && make install ; popd
+ln -s ~/go/bin/vcsim ~/bin # or /usr/local/bin
+ln -s ~/go/bin/govc ~/bin  # or /usr/local/bin
+```
+
+# creating simulator provider
+
+```bash
+echo "127.0.0.1 vc91" > /etc/hosts
```
agrare, line 20: Is this necessary?  You can just use `localhost` or `127.0.0.1` when adding the provider

(I know you were trying to avoid hostname collisions with >1 vcsim but that's also more advanced than the typical UI dev will need)

providers/vmware_simulator.md
```diff
@@ -0,0 +1,98 @@
...
+  govc metric.interval.change
+
+  # this one has a default (not sure if related or not)
+  govc object.collect -s -dump PerformanceManager:PerfMgr perfCounter
+```
+
+# run
+
+```bash
+# I dumped to a local directory to add historical capture support
```
Fryguy, line 96: Let's try to make the doc less of a brain dump and more of a guide to others.  I'm not sure what this is supposed to me as an outside reader.

providers/vmware_simulator.md
```diff
@@ -0,0 +1,98 @@
+# installing vmware simulator
```
Fryguy, line 1: I think we need an introduction of some sort.

```suggestion
# VMware Simulator

This guide explains general usage of the VMware simulator, `vcsim`.  This simulator can be used when developing the VMware provider.

## Installing VMware Simulator
```

Fryguy: Not the title case change at the end too.

providers/vmware_simulator.md
```diff
@@ -0,0 +1,98 @@
...
+git checkout miq_counters_kb
+# client
+pushd govc ; make clean && make && make install ; popd
+# server
+pushd vcsim ; make clean && make && make install ; popd
+ln -s ~/go/bin/vcsim ~/bin # or /usr/local/bin
+ln -s ~/go/bin/govc ~/bin  # or /usr/local/bin
+```
+
+# creating simulator provider
```
Fryguy, line 17:
```suggestion
## Running the simulator as a provider
```

providers/vmware_simulator.md
```diff
@@ -0,0 +1,98 @@
...
+      -host 2 -nsx 2 -pg 2 -pg-nsx 2 -pod 2 -pool 1 -standalone-host 2 -vm 2 \
+      -l vc91:9191 &
+
+export GOVC_URL=https://root:vmware@vc91:9191/sdk GOVC_INSECURE=true
+govc object.save -d my-vcenter
+# stop stop temporary simulator
+kill %1 # or fg ; control-c
+```
+
+# adding historicals
```
Fryguy, line 33: Thinking this should be another level deep under the "Running a simulator"

```suggestion
### Including historical metrics data
```

providers/vmware_simulator.md
```diff
@@ -0,0 +1,98 @@
...
+echo "127.0.0.1 vc91" > /etc/hosts
+
+# run temporary simulator 
+vcsim -username root -password vmware -app 2 -cluster 2 -dc 1 -ds 2 -folder 2 \
+      -host 2 -nsx 2 -pg 2 -pg-nsx 2 -pod 2 -pool 1 -standalone-host 2 -vm 2 \
+      -l vc91:9191 &
+
+export GOVC_URL=https://root:vmware@vc91:9191/sdk GOVC_INSECURE=true
+govc object.save -d my-vcenter
+# stop stop temporary simulator
```
Fryguy, line 29:
```suggestion
# stop temporary simulator
```

providers/vmware_simulator.md
```diff
@@ -0,0 +1,98 @@
...
+        <samplingPeriod>86400</samplingPeriod>
+        <length>31536000</length>
+        <name>Past Year</name>
+        <level>1</level>
+        <enabled>true</enabled>
+      </PerfInterval>
+    </val>
+  </propSet>
+```
+# better editing of historicals
```
Fryguy, line 78: I'm not sure what this title means - feels like this is part of the previous section,
```suggestion
```

providers/vmware_simulator.md
```diff
@@ -0,0 +1,98 @@
...
+
+TODO: please figure out a way to add historicals without editing the files.
```
Fryguy, line 80:
```suggestion
TODO: Figure out a way to add historicals without editing the files.
```

providers/vmware_simulator.md
```diff
@@ -0,0 +1,98 @@
...
+  govc object.collect -s -dump PerformanceManager:PerfMgr historicalInterval
+  govc metric.interval.info
+
+  govc metric.interval.change
+
+  # this one has a default (not sure if related or not)
+  govc object.collect -s -dump PerformanceManager:PerfMgr perfCounter
+```
+
+# run
```
Fryguy, line 93:
```suggestion
## Running the simulated provider from the generated data
```

providers/vmware_simulator.md
```diff
@@ -0,0 +1,98 @@
...
+
+# creating simulator provider
+
+```bash
+echo "127.0.0.1 vc91" > /etc/hosts
+
+# run temporary simulator 
+vcsim -username root -password vmware -app 2 -cluster 2 -dc 1 -ds 2 -folder 2 \
+      -host 2 -nsx 2 -pg 2 -pg-nsx 2 -pod 2 -pool 1 -standalone-host 2 -vm 2 \
+      -l vc91:9191 &
```
agrare, line 25: Why run in the backround?  Seems much simpler to leave it running in the foreground and let people ctrl-c to exit than telling them to do `kill %1` later

providers/vmware_simulator.md
```diff
@@ -0,0 +1,98 @@
...
+
+export GOVC_URL=https://root:vmware@vc91:9191/sdk GOVC_INSECURE=true
+govc object.save -d my-vcenter
```
Fryguy, line 28:
```suggestion
# dump simulator contents to a directory for later reuse
export GOVC_URL=https://root:vmware@vc91:9191/sdk GOVC_INSECURE=true
govc object.save -d my-vcenter
```

providers/vmware_simulator.md
```diff
@@ -0,0 +1,98 @@
...
+# stop stop temporary simulator
+kill %1 # or fg ; control-c
+```
+
+# adding historicals
+
+```xml
+<!--
+  edit my-vcenter/0009-PerformanceManager-PerfMgr.xml
+  find nistoricalInterval
```
Fryguy, line 38:
```suggestion
  find historicalInterval
```

providers/vmware_simulator.md
Fryguy, line 25: Not required, but I tend to find guides more readable if they are one param per line.  It also allows easier modification for things like removing lines.

```suggestion
vcsim \
   -username root \
   -password vmware \
   -app 2 \
   -cluster 2 \
   -dc 1 \
   -ds 2 \
   -folder 2 \
   -host 2 \
   -nsx 2 \
   -pg 2 \
   -pg-nsx 2 \
   -pod 2 \
   -pool 1 \
   -standalone-host 2 \
   -vm 2 \
   -l vc91:9191 &
```

Also is that `-|` at the end there  intentional?

agrare:
@Fryguy @kbrock high level I'm wondering if we're trying to show devs how to run a vmware simulator, or show them how to patch in @kbrock 's metrics fix?

This all would be drastically simpler and I think cover 90% of the use cases if we just say `go install ...` and `vcsim ...` and when his fix is merged upstream then we don't need to do anything to be able to support metrics.

providers/vmware_simulator.md
agrare: Right but that's only necessary so that you can hack in historical metrics, for normal vcsim running it isn't needed

Fryguy:
Actually now that you said that, I'm thinking this entire doc should probably be in the VMware repo in the Readme or in a docs dir, and then here maybe only a link to it from a "testing providers" doc or something.

Fryguy:
Oh yeah I saw that, but I read it as "leave the base one here and puts the metrics one over there", but I'm saying, "put the entire doc over there with the metrics too"

providers/vmware_simulator.md
```diff
@@ -0,0 +1,98 @@
+# installing vmware simulator
+
+```bash
+cd ~/go/
+git clone https://github.com/kbrock/govmomi
+cd govmomi
+git remote add upstream  https://github.com/vmware/govmomi
+git checkout miq_counters_kb
```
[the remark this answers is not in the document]
kbrock: agreed. we can now use master now that this is all merged

providers/vmware_simulator.md
kbrock: in the end of the day, vc91 (even if in `/etc/hosts`) was freaking out, so I just use localhost.

providers/vmware_simulator.md
kbrock: this is all moot now. vcsim out of the box properly creates those entries

kbrock:
WIP:
TODO: change to drop the editing of files, dumping of entries, and using non standard git location.

providers/vmware_simulator.md
kbrock: (my goal was to have multiple simulators running on the same machine. (one that represented a small ems and the other with a large ems)

We have issues with unique hostnames.
Instead, I started running these on unique ports

kbrock:
The vcsim currently works out of the box. The only extra thing we need to do to create a template.

PR #536, opened 2024-01-10
Add Discussions to triage

Fryguy:


triage_process.md
```diff
@@ -19,6 +19,8 @@ The ManageIQ core team holds regular meetings to triage issues and pull requests
   * [Unassigned PRs][]
   * [PRs without scope label][]
   * [Stale PRs][]
+* Discussions
+  * [Open discussions][]
```
jrafanie, line 23: Should we have a link for Unanswered Q&A also?  I think that would follow the pattern above where we assume triage will answer questions and close open discussions either by creating issues from them or asking for information that we'll review at the next triage.  At some point, if any question goes without reply, we'll close them.

jrafanie: or we can imply that open discussions will include both discussions and the special case q and a.

Fryguy: yeah I was thinking the latter at least for this initial pass.  If we feel breaking it up into separate line items is helpful we can do that in a follow up.

jrafanie: yeah, let's get this in and dog-food it.

PR #541, opened 2024-03-19
Add supports :create to writing_a_new_provider

agrare:
This should be created by the generator (https://github.com/ManageIQ/manageiq/pull/22950) but it should be included here as well.

### DIFF
--- a/providers/writing_a_new_provider.md
+++ b/providers/writing_a_new_provider.md
@@ -282,6 +282,9 @@ You basically define what forms you need in a hash in your provider plugin and t
 
 ```ruby
 class ManageIQ::Providers::AwesomeCloud::CloudManager < ManageIQ::Providers::CloudManager
+  supports :create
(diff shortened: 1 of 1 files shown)


providers/writing_a_new_provider.md
```diff
@@ -339,9 +342,7 @@ class ManageIQ::Providers::AwesomeCloud::CloudManager < ManageIQ::Providers::Clo
     ManageIQ::Providers::AwesomeCloud::Regions.all.map { |region| {:label => region[:name], :value => region[:name]} }
   end
 
-  supports :regions
   validates :provider_region, :inclusion => {:in => ManageIQ::Providers::AwesomeCloud::Regions.names}
-end
```
agrare, line 344: This isn't actually the entire class, there are a number of other methods built by the provider generator so removing the `end` here to not make it seem like this is all you need

PR #546, opened 2024-05-08
Add developer ovirt-engine-sdk instructions for C compiler errors

jrafanie:

### DIFF
--- a/developer_setup.md
+++ b/developer_setup.md
@@ -209,13 +209,27 @@ bin/setup
 
 ###### NOTE
 
-macOS requires platform specific Gems. Run `bundle config specific_platform true` before running `bin/setup`.
+- macOS requires platform specific Gems. Run `bundle config specific_platform true` before running `bin/setup`.
 
-If you've run PostgreSQL in a container, be sure to export the `DATABASE_URL` variable to connect to the container over TCP instead of a UNIX file socket.
+- The ovirt-engine-sdk gem contains C code that generated warnings with older Clang versions.  These warnings are marked as errors in newer Clang 15 and 16 depending on the Clang version, architecture, and possibly OS.  If this is the case, you can mark these as warnings and not errors to allow installation:
 
-```bash
-export DATABASE_URL='postgresql://localhost:5432' # optional, only necessary if PostgreSQL is running in a container
-```
+  ```
+  gem install ovirt-engine-sdk -- --with-cflags="-Wno-error=incompatible-function-pointer-types -Wno-error=implicit-function-declaration"
+  ```
+
+  This can also be configured within bundler:
+
+  ```
+  bundle config build.ovirt-engine-sdk --with-cflags="-Wno-error=incompatible-function-pointer-types -Wno-error=implicit-function-declaration"
+  ```
+
+  See https://github.com/oVirt/ovirt-engine-sdk-ruby/issues/14 and https://github.com/oVirt/ovirt-engine-sdk-ruby/issues/11.  Hopefully, the code in the gem can be resolved to avoid these warnings in the future.
+
+- If you've run PostgreSQL in a container, be sure to export the `DATABASE_URL` variable to connect to the container over TCP instead of a UNIX file socket.
+
+  ```bash
+  export DATABASE_URL='postgresql://localhost:5432' # optional, only necessary if PostgreSQL is running in a container
+  ```
 
 ## Start ManageIQ
 


developer_setup.md
```diff
@@ -211,6 +211,22 @@ bin/setup
...
+bundle config build.ovirt-engine-sdk --with-cflags="-Wno-error=incompatible-function-pointer-types -Wno-error=implicit-function-declaration"
+```
+
+See https://github.com/oVirt/ovirt-engine-sdk-ruby/issues/14 and https://github.com/oVirt/ovirt-engine-sdk-ruby/issues/11
+
+Hopefully, the code in the gem can be resolved to avoid these warnings in the future.
+
 If you've run PostgreSQL in a container, be sure to export the `DATABASE_URL` variable to connect to the container over TCP instead of a UNIX file socket.
 
 ```bash
```
Fryguy, line 232: This section is weird cause now it has 3 separate notes that run into each other, so it's hard to differentiate the individual problems...I'm thinking this should be a bullet (and the other 2 be bullets as well, but the GitHub UI won't let me do that in this suggestion - but I think you get the point).

```suggestion
- macOS requires platform specific Gems. Run `bundle config specific_platform true` before running `bin/setup`.

- The ovirt-engine-sdk gem contains C code that generated warnings with older CLANG versions.  These warnings are marked as errors in newer CLANG 15 and 16 depending on the CLANG version, architecture, and possibly OS.  If this is the case, you can mark these as warnings and not errors to allow installation:

  ```
  gem install ovirt-engine-sdk -v4.6.0 -- --with-cflags="-Wno-error=incompatible-function-pointer-types -Wno-error=implicit-function-declaration"
  ```

  This can also be configured within bundler:

  ```
  bundle config build.ovirt-engine-sdk --with-cflags="-Wno-error=incompatible-function-pointer-types -Wno-error=implicit-function-declaration"
  ```

  See https://github.com/oVirt/ovirt-engine-sdk-ruby/issues/14 and https://github.com/oVirt/ovirt-engine-sdk-ruby/issues/11

  Hopefully, the code in the gem can be resolved to avoid these warnings in the future.

- If you've run PostgreSQL in a container, be sure to export the `DATABASE_URL` variable to connect to the container over TCP instead of a UNIX file socket.

  ```bash
```

developer_setup.md
```diff
@@ -211,6 +211,22 @@ bin/setup
 
 macOS requires platform specific Gems. Run `bundle config specific_platform true` before running `bin/setup`.
 
+The ovirt-engine-sdk gem contains C code that generated warnings with older CLANG versions.  These warnings are marked as errors in newer CLANG 15 and 16 depending on the CLANG version, architecture, and possibly OS.  If this is the case, you can mark these as warnings and not errors to allow installation:
```
Fryguy, line 214:
```suggestion
The ovirt-engine-sdk gem contains C code that generated warnings with older Clang versions.  These warnings are marked as errors in newer Clang 15 and 16 depending on the Clang version, architecture, and possibly OS.  If this is the case, you can mark these as warnings and not errors to allow installation:
```

developer_setup.md
```diff
@@ -211,6 +211,22 @@ bin/setup
...
+
+```
+gem install ovirt-engine-sdk -v4.6.0 -- --with-cflags="-Wno-error=incompatible-function-pointer-types -Wno-error=implicit-function-declaration"
```
Fryguy, line 217: I'm wondering if we should remove the `-v4.6.0` here, since it could apply to future versions too and we don't want to have to keep updating this.  Besides, the application seems to use 4.4.1 anyway.

developer_setup.md
```diff
@@ -211,6 +211,22 @@ bin/setup
...
+
+This can also be configured within bundler:
+
+```
+bundle config build.ovirt-engine-sdk --with-cflags="-Wno-error=incompatible-function-pointer-types -Wno-error=implicit-function-declaration"
+```
+
+See https://github.com/oVirt/ovirt-engine-sdk-ruby/issues/14 and https://github.com/oVirt/ovirt-engine-sdk-ruby/issues/11
+
+Hopefully, the code in the gem can be resolved to avoid these warnings in the future.
```
Fryguy, line 228: Think we can merge these into 1 paragraph.
```suggestion
See https://github.com/oVirt/ovirt-engine-sdk-ruby/issues/14 and https://github.com/oVirt/ovirt-engine-sdk-ruby/issues/11.  Hopefully, the code in the gem can be resolved to avoid these warnings in the future.
```

jrafanie:
Ok, I believe I included all of the suggestions.

PR #549, opened 2024-07-18
Update openshift developer setup guide

agrare:
The instructions here are out of date and fail when followed using a current openshift cluster, specifically 
`ClusterRole` has been moved out of the root apiGroup and into `rbac.authorization.k8s.io/v1`, also an `apiGroups` property is required in `ClusterRole.rules` now

* Drop minishift instructions since we do not support openshift v3 any longer
* ~~TODO update CRC instructions to fix issues with current versions~~
* Point to the official openshift documentation as it is kept up-to-date on how to create cluster-roles / service-accounts

### DIFF
--- a/providers/openshift.md
+++ b/providers/openshift.md
@@ -108,31 +108,7 @@ vm openshift on your development machine) but is an OpenShift v4 cluster.
     oc login -u kubeadmin -p KUBEADMING_PASSWORD https://api.crc.testing:6443
     ```
 
-5. Now you can setup the management-infra/management-admin service account for use with ManageIQ
-    ```console
-    git clone https://gist.github.com/e2fac8be87ea0e9f429b6f5d75e02176 /tmp/manageiq_crc
-    oc adm new-project management-infra --description="Management-Infrastructure"
-    oc create serviceaccount management-admin -n management-infra
-    oc create -f /tmp/manageiq_crc/management-infra-admin-cluster-role.json
-    oc policy add-role-to-user -n management-infra admin -z management-admin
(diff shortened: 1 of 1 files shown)


providers/openshift.md
```diff
@@ -147,20 +71,20 @@ vm openshift on your development machine) but is an OpenShift v4 cluster.
...
+1. [Download](https://console.redhat.com/openshift/create/local) ane extract the latest release of CRC
 
     ```console
     tar xfJ crc-linux-amd64.tar.xz
     ```
 2. Copy the crc binary to a location on your PATH
 
     ```console
-    sudo cp crc-linux-1.8.0-amd64/crc /usr/local/bin
+    sudo install crc-linux-2.39.0-amd64/crc /usr/local/bin/crc
```
agrare, line 87: Using `install` will work even if crc is currently running (cp fails with a text file busy error)

providers/openshift.md
```diff
@@ -184,31 +108,7 @@ vm openshift on your development machine) but is an OpenShift v4 cluster.
     oc login -u kubeadmin -p KUBEADMING_PASSWORD https://api.crc.testing:6443
     ```
 
-5. Now you can setup the management-infra/management-admin service account for use with ManageIQ
-    ```console
-    git clone https://gist.github.com/e2fac8be87ea0e9f429b6f5d75e02176 /tmp/manageiq_crc
```
agrare, line 189: This gist was quite old and it is better to point to our existing openshift documentation since that is kept up to date

Fryguy (approved):

PR #553, opened 2024-11-13
[WIP] Use Rails credentials instead of deprecated and removed secrets

jrafanie:
Rails 7.1 removes access to modifying secrets as we should be moved over to rails credentials.  Here we describe how to setup these rails crednetials for the purpose of recording VCR cassettes.

See also: https://github.com/ManageIQ/manageiq-providers-autosde/pull/253

### DIFF
--- a/providers/writing_vcr_specs.md
+++ b/providers/writing_vcr_specs.md
@@ -22,33 +22,70 @@ The next thing we have to take care of is hiding "secrets".  Since the VCR YAML 
 
 VCR handles this with the `config.define_cassette_placeholder` option.  You provide VCR with a string that you want to be replaced, and then what you want it to be replaced with.  This allows for hostnames / passwords / etc... to be used when recording the cassette but the values will not be written to the resulting YAML files.
 
-ManageIQ has a pattern to help you with this, simply create a `config/secrets.defaults.yml` file:
-```yaml
----
-test:
-  awesome_cloud_defaults: &awesome_cloud_defaults
-    access_key: AWESOME_CLOUD_ACCESS_KEY
-    secret_key: AWESOME_CLOUD_SECRET_KEY
-  awesome_cloud:
-    <<: *awesome_cloud_defaults
+ManageIQ has a pattern to help you with this.  We use rails credentials.  Run the following command in the main application directory:
+
+```
+EDITOR=vi be rails credentials:edit --help
 ```
 
-Then create a `config/secrets.yml` file (this file will not be committed and should be in your .gitignore):
+This will provide more information about rails credentials.
+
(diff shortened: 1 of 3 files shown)


providers/writing_vcr_specs.md
```diff
@@ -22,34 +22,71 @@ The next thing we have to take care of is hiding "secrets".  Since the VCR YAML
...
+
+Add memoized methods in spec/spec_helper.rb with the defaults you want to assumed when running from cassettes to avoid leaking actual credentials:
+
+```ruby
+def credentials_awesome_cloud_host
+  @credentials_awesome_cloud_host ||= Rails.application.credentials.awesome_cloud_host || "awesome-cloud-host"
+end
+
+def credentials_awesome_cloud_user
+  @credentials_awesome_cloud_user ||= Rails.application.credentials.awesome_cloud_user || "awesome-cloud-user"
```
agrare, line 64: Should this be `Rails.application.credentials.awesome_cloud&.access_key`

providers/writing_vcr_specs.md
```diff
@@ -22,34 +22,71 @@ The next thing we have to take care of is hiding "secrets".  Since the VCR YAML
...
 
 Then add the following to your `VCR.configure` block in `spec/spec_helper.rb` after setting the `config.cassette_library_dir`:
+
+
 ```ruby
-  secrets = Rails.application.secrets
-  secrets.awesome_cloud.each do |key, val|
-    config.define_cassette_placeholder(secrets.awesome_cloud_defaults[key]) { val }
+  defaults = {
+    "host_key"   => credentials_awesome_cloud_host,
```
agrare, line 77:
```suggestion
```

providers/writing_vcr_specs.md
jrafanie: Thanks, I'm trying to generalize them.  Does that make sense for a provider that uses user/password authentication?  If so, I can do that.

agrare: Well two things, one I think even if there was a `user: "..."` it would be `Rails.application.credentials.awesome_cloud.user` not `Rails.application.credentials.awesome_cloud_user` and two the example here has `access_key` not `user`

I don't think I'd want to change the "writing a provider" example to have host/user/password because that doesn't match with the rest of the example code (there's an example connect method and an example api_client, etc...)

agrare: Maybe in addition to editing this specific example to work with the new credentials interface we need a "recording_vcrs.md" guide that covers how to do this more generally?

jrafanie:
Closing in favor of https://github.com/ManageIQ/manageiq/pull/23496 since Rails credentials is geared more towards teams sharing encrypted credentials, which is not what we need for VCR secrets.

PR #561, opened 2025-04-09
Fix minor issues with IPA docs

Fryguy:
@jrafanie Please review.

NTP settings in the UI no longer exists, which is why that section removes it.


jrafanie (approved):
LGTM

PR #563, opened 2025-05-02
Update node version in docs to Node 20

GilbertCherrie:
Part of https://github.com/ManageIQ/manageiq/issues/23425

Update the minimum Nodejs version in the docs to Node 20.19.1


GilbertCherrie:
@miq-bot add-label @Fryguy

PR #564, opened 2025-05-08
Fixing typo in oidc authentication.conf

ahrechushkin:
Removed extra % symbol in authentication.conf

### DIFF
--- a/external_auth/oidc-httpd-configs/authentication.conf
+++ b/external_auth/oidc-httpd-configs/authentication.conf
@@ -36,7 +36,7 @@ OIDCOAuthSSLValidateServer         Off
   FileETag                   None
   Header Set Cache-Control   "max-age=0, no-store, no-cache, must-revalidate"
   Header Set Pragma          "no-cache"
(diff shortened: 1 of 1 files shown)


Fryguy (approved):
Note that this is only the guides. The actual shipped changes need to be made in 

- (Needs a fix) https://github.com/ManageIQ/manageiq-pods/blob/master/manageiq-operator/api/v1alpha1/helpers/miq-components/httpd_conf.go#L274
- (Looks ok) https://github.com/ManageIQ/manageiq-appliance/blob/master/TEMPLATE/etc/httpd/conf.d/manageiq-external-auth-openidc.conf.erb#L33

@ahrechushkin Can you also make a PR to the manageiq-pods repo for the first one?

cc @bdunne

ahrechushkin:
> Note that this is only the guides. The actual shipped changes need to be made in

@Fryguy, to be honest, the external authentication guide does not contain any links to the ManageIQ appliance templates. Perhaps it would make sense to add links to the actual configuration files?

For your information, I believe this pull request and this issue are related: [https://github.com/orgs/ManageIQ/discussions/21893 ](https://github.com/orgs/ManageIQ/discussions/21893).

PR #565, opened 2025-06-27
[WIP] Add cypress set up documentation

GilbertCherrie:
Add cypress set up documentation.

PR #599, opened 2026-04-17
Bring over cypress guide from ui-classic

jrafanie:
There was existing cypress command and assertion reference information in the UI classic cypress readme.  I wanted to include Gilbert's guide in #565, so I merged that content with what existed already and cleaned up some things.  That work was done in https://github.com/ManageIQ/manageiq-ui-classic/pull/9988

Now that the content has been merged and cleaned up a bit, we're bringing all that content back to guides in this PR and UI classic can link to it.

All that for this!

Followup PR to update ui classic to link to the new guide here: https://github.com/ManageIQ/manageiq-ui-classic/pull/9990

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@
   - [Interactive debugging with Pry-Remote](developer_setup/debugging.md)
 * [Running in minimal mode](developer_setup/minimal_mode.md)
 * [Running the test suites](developer_setup/running_test_suites.md)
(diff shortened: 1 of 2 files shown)


Fryguy:
I think this might fit better on the UI subdir.

jrafanie:
> I think this might fit better on the UI subdir.

fixed

Fryguy (approved):

PR #600, opened 2026-05-19
Bump Ruby minimum to 3.3.1

Fryguy:
zeitwerk 2.8.0 causes an error on Ruby 3.3.0 which is fixed in 3.3.1, so we are bumping the minimum

@jrafanie @GilbertCherrie Please review.

Depends on
- https://github.com/ManageIQ/manageiq/pull/23848

### DIFF
--- a/developer_setup.md
+++ b/developer_setup.md
@@ -4,7 +4,7 @@
 
 | **Name**   | **Min Version** | **Max Version** |
 | ---------- | --------------- | --------------- |
-| Ruby       | 3.3.x           | 3.4.x           |
+| Ruby       | 3.3.1           | 3.4.x           |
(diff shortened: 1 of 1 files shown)


GilbertCherrie (approved):

jrafanie (approved):
Looks good to me!  😎  👍

jrafanie:
Merging, but it's worth pointing out that zeitwerk was released to fix this issue with 3.3.0, see https://www.github.com/fxn/zeitwerk/issues/334

> I have released 2.8.1 removing the syntax (valid since Ruby 3.1) that 3.3.0 is not able to parse.

jrafanie (approved):
LGTM

